PR #2, opened 2026-06-30
Ship duckling gem: time extraction via Magnus + wafer-inc-duckling

cpb:
## Ship duckling gem: time extraction via Magnus + wafer-inc-duckling

Implements `Duckling.parse`, closing #1. The gem now compiles a native Rust
extension (via [Magnus](https://github.com/matsadler/magnus) + `rb_sys`) that
links the [`duckling`](https://crates.io/crates/duckling) crate
(wafer-inc's Rust port of Facebook's Duckling) and extracts time/date
entities from English text — no Haskell service, no HTTP call.

## API

```ruby
Duckling.parse(text, locale: "en", dims: ["time"], reference_time: nil, with_latent: false)
# => [{ body: "tomorrow", start: 8, end: 16, dim: :time, latent: false,
#       value: { type: :value, value: "2013-02-13T00:00:00", grain: :day,
#                 values: [...] } }]
```

- `locale:` — BCP-47 tag (`"en"`, `"en-GB"`); unsupported codes raise `ArgumentError`.
- `dims:` — dimension names to extract; defaults to `["time"]`. Only `:time` is
  wired up in 0.2.0 — other dimension strings raise `ArgumentError`.
- `reference_time:` — Unix seconds; anchors relative expressions like
  "tomorrow" for deterministic tests. Defaults to `Time.now`.
- `with_latent:` — include ambiguous/latent matches (e.g. "morning" alone).
- Interval expressions ("from 3pm to 5pm") return `type: :interval` with
  `:from`/`:to` sub-hashes instead of a flat `:value`.

## What changed

- `ext/duckling/{Cargo.toml,extconf.rb,src/lib.rs}` — new cdylib extension
  crate. Entity/TimeValue/TimePoint/Grain are mapped to Ruby by hand
  (`RHash`/`RArray`), not via `serde_magnus` — the crate's derived serde
  shapes (externally-tagged enums, PascalCase grains) don't match the target
  Ruby wire format.
- `Rakefile` — `Rake::ExtensionTask` wired in; `rake compile` now runs before
  `test`.
- `.github/workflows/main.yml` — added a Rust toolchain step.
- `lib/duckling.rb` — now just loads the compiled extension.
- `lib/duckling/version.rb` — bumped to `0.2.0`.

## Known limitations (0.2.0, tracked for follow-up)

- `reference_time:` is reconstructed as UTC+0 internally, which loses the
  caller's original offset — fine for the current test corpus, worth
  revisiting for exact-offset assertions later.
[...]

Issue #1, opened 2026-06-30
Ship duckling gem: time extraction via Magnus + wafer-inc-duckling


cpb:
## Goal

Ship a Ruby gem that extracts named entities from text at recognition time by
wrapping the wafer-inc-duckling Rust crate via Magnus — no Haskell runtime,
no HTTP call.

## Background

The existing duckling ecosystem offers two Ruby-hostile paths: the original
Haskell service (requires HTTP at recognition time) and pyduckling (Python FFI,
not usable from Ruby). wafer-inc-duckling is a Rust-native port of the
duckling entity parser; Magnus makes it practical to wrap Rust crates as native
Ruby extensions. This project combines them to bring duckling-quality entity
extraction to Ruby, beginning with time/date entities for a 0.2.0 release,
then iterating on additional dimensions and locales.

## Acceptance criteria

- [x] The `duckling` gem compiles a native extension linking the
      wafer-inc-duckling Rust crate via Magnus
- [x] `Duckling.parse(text, locale: "en")` extracts time/date entities from
      English text and returns structured results matching what
      wafer-inc-duckling produces for the same input
- [x] Test suite includes cases drawn from wafer-inc-duckling and pyduckling
      test suites to validate parity
- [ ] `0.2.0` is released through the automated pipeline established in issue #4
- [ ] CI runs on push: minitest tests + Standard linter pass

## Out of scope

- Non-time entity dimensions (number, amount-of-money, distance, etc.) — subsequent releases
- Non-English locales — planned after time extraction is stable (language parity before entity parity)
- Robust regression suite — deferred until after 0.2.0 ships

## Notes

- The `research-pr` skill will produce terrain-map documents covering gem
  release workflow for Magnus crates, API design, repo organization, and
  end-to-end integration approach; hill-first PRs will emerge from that output
- `ext/duckling/extconf.rb` is scaffolded but empty — the research phase will
  determine the right build wiring (Cargo-based vs. rb-sys)
- 0.1.0 already published to RubyGems
- Issue #4 must ship first (establishes the release pipeline used here)

cpb:
v0.2.0 shipped. All hill layers merged. Closing.

PR #3, opened 2026-06-30
research(1): ship duckling gem — time extraction via Magnus + wafer-inc-duckling

cpb:
Research and planning artifacts for issue #1. Entry point: docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/README.md

## What this PR answers

How to ship a Ruby gem that extracts time/date entities from text using the wafer-inc-duckling Rust crate wrapped via Magnus — no Haskell runtime, no HTTP call — and publish it as `duckling` 0.2.0.

## Document tree

```
docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/
  README.md                       ← entry point, reading order, settled decisions, open questions
  research/
    README.md                     ← breadth-first summary, key findings table
    wafer-inc-duckling-api/       ← complete Rust API surface, Naive vs Instant semantics
    build-wiring/                 ← Magnus + rb-sys build plumbing, CI wiring
    type-mapping-strategy/        ← serde analysis → manual Magnus mapping; symbol key decision
    test-coverage/                ← corpus cases, Ruby minitest design, pyduckling parity scope
    ffi-risks.md                  ← empirical evaluation of GVL blocking, panic safety, date rot, day-of-week gap
  plans/
    README.md                     ← dependency order, key decisions table
    00-pr2-roadmap.md             ← step-by-step path from research → PR #2 green ← START HERE
    01-native-extension-setup.md  ← wire cdylib crate, extconf.rb, Rakefile, CI
    02-ruby-api-design.md         ← Duckling.parse API, manual Magnus mapping, symbol keys, NaiveDateTime handling
    03-test-suite-and-ci.md       ← hill tests already written; extended corpus design
```

## Settled decisions

1. **duckling 0.4.0 on crates.io** — use `duckling = "0.4"` in Cargo.toml. No publish blocker.
2. **Symbol keys and values** — all entity hash keys (`:body`, `:dim`, `:value`, `:type`, `:grain`) and grain/type/dim values are Ruby Symbols. Settled by PR #2's hill tests.
3. **NaiveDateTime → bare ISO8601** — no offset appended to Naive datetimes. Semantically honest; hill tests use prefix matching.
4. **Manual Magnus mapping** — `serde_magnus` produces wrong enum shapes. Option B (manual `RHash`/`RArray` construction) is required.
5. **GVL hold accepted for 0.2.0** — Magnus 0.9 has no high-level GVL release API (`rb_thread_call_without_gvl` is not wrapped). Measured parse time is ~500µs for short inputs; acceptable at low concurrency. Benchmark deferred to 0.2.1.
6. **Panics already mitigated** — wafer-inc-duckling wraps the parser in `catch_unwind` in release builds. Bridge code must use `?` not `.unwrap()`; add a `catch_unwind` wrapper around the bridge function as belt-and-suspenders.

## Open questions (still requiring human decision or verification)

1. **`reference_time:` timezone loss** — passing `reference_time` as a Unix `i64` reconstructs at UTC+0, losing the caller's offset. The hill tests don't expose this, but extended corpus tests for `"now"` will require exact ISO8601 with offset. Option: accept Ruby `Time` and extract `.utc_offset` via Magnus in 0.3.0.

2. **`NoGrain` string** — `Grain::as_str()` returns `"no_grain"`; pyduckling uses `"nosec"`. Verify whether any real `Time` entity in the corpus actually carries `NoGrain` before shipping.

3. **Day-of-week validation gap** — duckling resolves the date from "Friday July 1st 2026" but does **not** flag that Friday is wrong (July 1, 2026 is Wednesday). This is a semantic gap in duckling's design (not a missing feature). The gem alone cannot detect day-of-week hallucinations; callers must compute the day independently and compare. Document in README.


docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/wafer-inc-duckling-api/types.md
```diff
@@ -0,0 +1,321 @@
...
+---
+
+## `TimeValue` — top-level time result
+
+```rust
+#[derive(Debug, Clone, PartialEq, serde::Serialize)]
+pub enum TimeValue {
+    Single {
+        value: TimePoint,
+        values: Vec<TimePoint>,      // up to 3 next occurrences, primary is first
```
cpb: **polish:** are there examples already in the docs for what these occurrences are? IE: christmas, and then the next 3 years? What other invocations result in occurrences?

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/wafer-inc-duckling-api/types.md
```diff
@@ -0,0 +1,321 @@
+# Public Types
+
+Source files:
+- `/Users/cpb/projects/duks/wafer-inc-duckling/src/types.rs`
+- `/Users/cpb/projects/duks/wafer-inc-duckling/src/dimensions/time_grain/mod.rs`
```
cpb: **issue:** remove all path references to my local file system and replace them with github permalinks to the read versions

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/README.md
```diff
@@ -0,0 +1,13 @@
+# Build Wiring Research
+
+How to wire up the native extension so the gem compiles: cdylib crate layout,
+extconf.rb content, Rakefile changes, and CI configuration.
+
+## Documents
+
+| File | Description |
+|------|-------------|
+| [extension-crate.md](extension-crate.md) | The new `ext/duckling/` Rust crate that must be created: cdylib type, Cargo.toml shape, `build.rs` question, and open dependency questions for wafer-inc-duckling. |
```
cpb: **issue:** any instance of `wafer-inc-duckling` in a markdown file must be formatted as in the readme, a link to their duckling repo labelled duckling

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/wafer-inc-duckling-api/types.md
cpb: Fixed in 33bd68d — replaced every `/Users/cpb/projects/duks/...` local path in the doc tree (13 files) with GitHub permalinks pinned to the commit actually examined:

- wafer-inc/duckling@c96b068
- matsadler/magnus@4e46772
- cpb/pyduckling@bbe06d2 (fork, upstream is treble-ai/pyduckling)
- cpb/duckling@ec00708 (for the Rakefile, which lives in this repo)

One exception: ci-configuration.md cited a `rust-toolchain.toml` in the local magnus checkout, but that file turned out to be untracked (`git status` confirms it, and it's absent from matsadler/magnus on GitHub) — so no valid permalink exists. I corrected that section to say Magnus doesn't actually pin a toolchain version upstream, rather than link to a file that isn't really there.

While verifying links I also caught a separate, pre-existing bug: several docs referenced `ruby.sym("body")` / `Ruby::sym` as the Magnus API for creating symbols. That method doesn't exist — magnus only has `ruby.to_symbol()` (defined in `src/symbol.rs`; there's no `src/ruby.rs` at all). Fixed across all 5 affected files.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/wafer-inc-duckling-api/types.md
cpb: Good catch — there weren't any examples. Added a new subsection ("What populates values (occurrences)") in 33bd68d that traces the actual logic instead of guessing:

- `values` is filled by the series generator (`generate_extra_values` / `series::generate_series` in wafer-inc-duckling), which only produces multiple entries for **recurring/unanchored** time forms: bare weekday names, bare month names, hour/time-of-day alone, day-of-month alone, `Year`, part-of-day, weekend, season, `Holiday` *without* an explicit year, and month/day-without-year dates — plus `Composed` combinations of these.
- Your "christmas" example is exactly right: unpinned "christmas" resolves as `Holiday(name, year_opt: None)` and walks forward year by year, giving `values: [<this year>, <next year>, <year after>]`. But "christmas 2026" (year pinned) short-circuits to a single value — no extra occurrences.
- A fully anchored date like "July 1st 2026" also only ever produces 1 value, though via a slightly different code path (the series returns exactly one object rather than an empty one that falls back to the primary).

No corpus test currently exercises the `values` field at all — the corpus check helpers explicitly ignore it with `..` — so this is derived from reading the resolution code directly, not from an existing worked example.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/README.md
cpb: **polish:** can you make all tables of documents link to their file, but have the text of the link their heading?

cpb: Fixed in be33945 — every bare "wafer-inc-duckling" prose mention across the doc tree (16 files, ~65 instances) is now `[duckling](https://github.com/wafer-inc/duckling)`, matching the format already used in research/wafer-inc-duckling-api/README.md.

Left unlinked, deliberately, matching that same reference README's own convention:
- Document headings/titles (the reference README doesn't link its own H1 either)
- Internal relative doc links whose link text is itself a folder/file path (e.g. `[wafer-inc-duckling-api/](...)`)
- Mentions inside Rust/TOML code fences — can't embed markdown link syntax in code

While in there I also caught and fixed a stale claim in 01-native-extension-setup.md saying the extension crate uses a local Cargo path dependency on wafer-inc-duckling — it's actually a crates.io dependency (`duckling = "0.4"`), as already settled elsewhere in the doc tree.

cpb: Fixed in 9c9b436 — there were exactly two markdown tables of documents (research/wafer-inc-duckling-api/README.md and research/build-wiring/README.md); both now use each linked file's H1 heading as the link text instead of the filename:

- `[public-functions.md]` → `[Public Parse Functions]`
- `[types.md]` → `[Public Types]`
- `[locale-system.md]` → `[Locale System]`
- `[extension-crate.md]` → `[Extension Crate Layout]`
- `[extconf-rb.md]` → `[extconf.rb Pattern]`
- `[rakefile-setup.md]` → `[Rakefile Setup]`
- `[ci-configuration.md]` → `[CI Configuration]`

The other "Documents"/"Table of Contents" sections in the doc tree (test-coverage, type-mapping-strategy, plans, top-level README) use bulleted lists rather than tables, so they're unaffected — let me know if you'd like those converted to tables too.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/extconf-rb.md
```diff
@@ -0,0 +1,77 @@
...
+spec.extensions = ["ext/duckling/extconf.rb"]
+spec.add_dependency "rb_sys", "~> 0.9.39"
+```
+
+The `rb_sys` runtime dependency is what provides `rb_sys/mkmf` to end-users
+who install the gem from source. Gemfile.lock pins `rb_sys` to `0.9.128`.
+
+## Open Questions
+
+1. **Does `create_rust_makefile` need Cargo in PATH at gem install time?**
```
cpb: **polish:** create an issue to ensure that `bin/setup` on macos uses homebrew to install any non-gem dependencies to be able to operate this project; likewise create an issue to ensure that bin/claude-code-web-setup has the needed dependencies just in time; create an issue to ensure that our ci targets the versions of ruby and rust tooling available without further installation in claude code web (check out this profile (https://github.com/cpb/fizzbuzz_app/blob/babd9ee74ccd2160561e3e74f53d3f43cd9ec848/doc/observation/2026-06-07-claude-code-web-environment.md); create a separate issue to expand testing to a matrix of: what claude code web environment supports, and ruby 3.4, the latest ruby, and the latest rust

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/extconf-rb.md
```diff
@@ -0,0 +1,77 @@
...
+   Yes — the Makefile it generates invokes `cargo build`. End-users installing
+   from source must have Rust/Cargo installed. Pre-compiled binary gems (see
+   ci-configuration.md) bypass this requirement.
+
+2. **Does rb_sys/mkmf need to be required before mkmf?** The rust_blank example
+   requires `mkmf` first, then `rb_sys/mkmf`. This ordering should be followed.
```
cpb: **polish:** are these questions still open given we appear to be building and releasing the gem?

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/rakefile-setup.md
```diff
@@ -0,0 +1,131 @@
...
+```ruby
+task :dev do
+  ENV["RB_SYS_CARGO_PROFILE"] = "dev"
+end
+```
+
+Used as: `bundle exec rake dev compile test`
+
+Without this, cargo defaults to the release profile (slower compile, faster
+runtime). During development the dev profile is preferred.
```
cpb: **polish:** create an issue to make this happen

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/rakefile-setup.md
```diff
@@ -0,0 +1,131 @@
...
+pins it to `1.2.9`. `Rake::ExtensionTask` is provided by this gem.
+
+## Open Questions
+
+1. **Does Minitest::TestTask.create support deps?** The current Rakefile uses
+   `Minitest::TestTask.create` (minitest 6.0.6). The rust_blank example uses
+   `Rake::TestTask` with `t.deps << :compile`. These are different classes.
+   Check whether `Minitest::TestTask.create` accepts a block with deps, or
+   whether the compile task needs to be listed explicitly in the default task
+   ordering.
```
cpb: **praise:** good idea, create an issue to explore this

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/rakefile-setup.md
```diff
@@ -0,0 +1,131 @@
...
+
+2. **Does `task default: %i[compile test standard]` guarantee compile runs
+   before test?** Rake executes prerequisite tasks in left-to-right order when
+   they are listed in the array, so `compile` running before `test` should hold.
+   Verify against actual Rake behavior with the installed rake version (13.4.2).
```
cpb: **nitpick:** my intuition, and I believe our shared experience, is that this holds true. Maybe do a sanity check and drop the open question?

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/ci-configuration.md
```diff
@@ -0,0 +1,215 @@
...
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+      - name: Run the default task
+        run: bundle exec rake
+```
+
+Missing from this workflow:
+- No Rust toolchain installation
+- No compile step — `bundle exec rake` currently runs `test` + `standard` only
+  (the default task does not include `compile`)
```
cpb: **issue:** this is now fully stale, I want you to re-draft this document before I read it please

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/README.md
```diff
@@ -0,0 +1,9 @@
+# Type Mapping Strategy
+
+Research into converting [duckling](https://github.com/wafer-inc/duckling)'s Rust types to Ruby values via Magnus.
+
+## Documents
+
+- [serialization-options.md](./serialization-options.md) — Compares the three conversion approaches (serde_magnus, manual Magnus mapping, JSON round-trip) with verified serde attribute analysis and a recommendation.
+- [ruby-hash-schema.md](./ruby-hash-schema.md) — Defines the target Ruby Hash shape for time entities in the 0.2.0 release, documenting the pyduckling compatibility target and the open NaiveDateTime timezone question.
+- [magnus-type-conversions.md](./magnus-type-conversions.md) — Documents Magnus 0.9.0's built-in Rust-to-Ruby type conversions, the chrono feature gate, and example Rust code for manually building the time entity hash.
```
cpb: **issue:** this file, and all README files at this depth need to have this format instead: docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/README.md -- we should never have the document link text be the file name, it should be the heading of that file. I do prefer the table format too.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/serialization-options.md
```diff
@@ -0,0 +1,184 @@
...
+must be known. Here is what the source shows:
+
+**`Entity`** (`#[derive(serde::Serialize)]`, no container attributes):
+- Field-level: `#[serde(skip_serializing_if = "Option::is_none")]` on `latent`
+- Struct fields serialize with their Rust names: `body`, `start`, `end`, `value`, `latent`
+
+**`DimensionValue`** (`#[derive(serde::Serialize)]`, no container attributes):
+- No `#[serde(tag = ...)]`, no `#[serde(rename_all = ...)]`
+- Default serde representation for enums with data: externally tagged
+- `DimensionValue::Time(tv)` serializes as `{"Time": <tv>}`, not `{"type": "time", ...}`
```
cpb: **suggestion:** An alternative handling I don't see considered is to:

lean into the Serde representations `"Time": <tv>` and leverage ruby pattern matching in factory methods for Ruby Data representations of the types. We can iterate on the ergonomics from there, in ruby … I do not prefer hashes.

Ruby pattern matching relies on symbol keys, so the only tweak to serde we should make is to symbolize the keys

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/README.md
```diff
@@ -0,0 +1,9 @@
+# Type Mapping Strategy
```
cpb: **note:** I really glossed over this because I feel like it is too anchored in pyduckling's decisions, and hash primacy. I have other comments about preferred directions. Will want to pursue that refactoring soon.

I'm also holding onto a todo for v0.3.0 about handling naive times the way rails activesupport does -- in a prior session it was suggested that is in this plan. use this prompt as a chance to double check that, and ensure we have a plan for a ticket in this branch for that (and, adopting serde_magnus verbatim symbol key output with ruby pattern matching factory construction of `Data` types)

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/extconf-rb.md
cpb: Filed as four separate issues (this is follow-on work beyond this docs-only PR, not something to implement here):

- #26 — `bin/setup` on macOS: maintain a `Brewfile` + `brew bundle` step for non-gem deps (starting with Rust)
- #27 — Fill in the no-op `bin/claude-code-web-setup` skeleton with just-in-time dependency install (bundle install now, native-extension compile once it lands), using the receipt-file caching pattern the skeleton already references
- #28 — Pin CI's Rust toolchain to what's pre-installed in Claude Code Web, instead of floating on the `stable` channel
- #29 — Expand the CI matrix to also cover Ruby 3.4, latest Ruby, and latest Rust (forward-compat signal beyond the Claude Code Web baseline)

One correction while researching #28: this branch was behind `main`, which already merged a Rust CI step (`dtolnay/rust-toolchain@stable`, pinned by commit SHA) plus `cargo fmt`/`clippy` checks against `ext/duckling/` — the native extension has landed on `main`. I didn't rebase this branch onto `main` (a force-push risk given other agents may be concurrently pushing to it per project convention) — I fetched and read `origin/main` directly to get accurate facts for the issue instead. Ruby's version (3.3.6) already matches Claude Code Web's active version; the real gap for #28 is that CI's Rust version floats on `stable` rather than being pinned to the sandbox's actual `1.94.1`.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/test-coverage/ruby-test-design.md
```diff
@@ -0,0 +1,496 @@
+# Ruby Minitest Test Suite Design
```
cpb: **praise:** great doc, can you make an issue to bring our tests up to date with this plan

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/test-coverage/pyduckling-reference.md
```diff
@@ -0,0 +1,252 @@
...
+
+This confirms the [duckling](https://github.com/wafer-inc/duckling) test suite was authored by comparing against the Haskell corpus output. The intent is parity with Haskell.
+
+The comment at line 1 of `time_en.rs`:
+> `/// English time training corpus, ported from Duckling/Time/EN/Corpus.hs.`
+
+### Practical implications for Ruby tests
+
+- For simple cases (today, tomorrow, at 3pm, in 2 hours), [duckling](https://github.com/wafer-inc/duckling) and pyduckling should return identical results.
+- The Ruby acceptance criteria (`pr_context.md`) explicitly says: "matching what [duckling](https://github.com/wafer-inc/duckling) produces" — so Ruby tests compare against the Rust output, not the Haskell/pyduckling output.
```
cpb: **nitpick:** don't refer to ephemeral files like `pr_context.md` in.

**polish:** in the future, would love to leverage the pyduckling checkout, or {rasa,facebook}/duckling 's tests to identify any gaps in wafer's and upstream fixes.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/extconf-rb.md
cpb: No — confirmed stale. Checked main directly: the extension shipped and v0.2.0 is already tagged/released (v0.1.0 → v0.1.1 → v0.1.2 → v0.2.0). Both open questions are answered by the shipped code:

1. Cargo in PATH at install time — yes, confirmed necessary (gem ships source-only, no pre-compiled binaries).
2. `mkmf` before `rb_sys/mkmf` ordering — confirmed exactly matching `ext/duckling/extconf.rb` on main.

Updated in 65f7af9: converted the "Open Questions" section to "Resolved," citing the actual shipped file.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/rakefile-setup.md
cpb: Filed as issue #30 — added a `:dev` Rake task setting `RB_SYS_CARGO_PROFILE=dev` for faster local edit-compile-test loops. Documented in rakefile-setup.md (65f7af9).

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/rakefile-setup.md
cpb: Filed as issue #31 — exploring whether `test` should explicitly depend on `:compile` (rather than relying only on the `default` task's array ordering), so `bundle exec rake test` run in isolation still compiles first. Documented in rakefile-setup.md (65f7af9).

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/rakefile-setup.md
cpb: Confirmed empirically rather than just sanity-checked by intuition — ran the actual installed Rake 13.4.2 with a tiny test Rakefile (three tasks with side-effecting puts calls listed as `task default: %i[c a b]`) and invoked it: they ran in exactly that order, c → a → b. Rake does guarantee prerequisite arrays run left-to-right. Dropped the open question in 65f7af9 and replaced it with this as a settled, verified fact.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/build-wiring/ci-configuration.md
cpb: Re-drafted in 65f7af9. Checked main directly rather than guessing: the extension has shipped, v0.2.0 is already tagged/released, and CI already has a Rust step (which this doc's original "what needs to change" framing didn't anticipate at all). The rewrite:

- Documents the actual current `.github/workflows/main.yml` and the new `release.yml` (tag-triggered, source-gem publish, auto-changelog PR) — neither existed when this doc was first drafted
- Confirms the source-gem-only decision shipped as predicted, cross-compilation was never built
- Flags that `ext/duckling/Cargo.toml` actually uses `magnus = "0.8"`, not `"0.9"` as several other research docs assumed (0.9 was never published to crates.io)
- Replaces "Open Questions" with either resolved facts or links to tracked issues (#28 Rust version pinning, #29 matrix expansion) instead of leaving them as open prose

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/README.md
cpb: Applied to type-mapping-strategy/README.md and test-coverage/README.md in 65f7af9 — both converted from bullet lists to the table + heading-link-text format matching build-wiring/README.md and wafer-inc-duckling-api/README.md. That's now the format for every subtree README's document table in this tree.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/serialization-options.md
cpb: Documented as "Option D" in serialization-options.md (65f7af9), written to preserve the idea rather than declare it decided: keep serde_magnus's externally-tagged shape as-is (don't fight it with rename_all/tag container attributes like Option A considered), symbolize the keys (the one tweak that matters for `case/in` pattern matching), and write Ruby factory methods that pattern-match on that shape to construct `Data`-based value objects instead of returning raw Hashes.

Filed as issue #32 to actually prototype and evaluate it — not adopted for 0.2.0 (which already shipped on Option B / manual mapping), but tracked as the concrete next step for whenever this direction gets picked up.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/README.md
cpb: Checked — the ActiveSupport-style naive-time TODO wasn't actually in this branch's plans yet (searched for "ActiveSupport"/"TimeWithZone" across the whole doc tree, found nothing). The closest existing thing was ruby-hash-schema.md's "Option N2: apply reference timezone at serialization time (deferred to 0.3.0)" — a one-line note, not a tracked ticket, and not framed as ActiveSupport-style.

Filed as issue #33, framed explicitly as you described: resolve Naive time values against the reference context's zone (ActiveSupport `Time.zone`-style) rather than surfacing them zoneless, for v0.3.0. Linked from both ruby-hash-schema.md and serialization-options.md (65f7af9).

Also filed #32 for the serde_magnus (symbol keys) + Ruby pattern-matching `Data` factory direction from your other comment — noted in both issues that they should be designed together since the naive-time value needs to fit whatever object shape (Hash vs. Data) is current when it's implemented. Added a note at the top of type-mapping-strategy/README.md flagging that the current recommendation (Option B, shipped for 0.2.0) is anchored in pyduckling's Hash-based decisions and that this preferred-direction shift is tracked separately, not silently folded into the existing recommendation.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
```diff
@@ -0,0 +1,405 @@
...
+and [matsadler/magnus@4e46772](https://github.com/matsadler/magnus/tree/4e46772050e47cd6cd988fa935263cc5c583e388).
+
+---
+
+## Summary Table
+
+| Risk | Hypothesis | Verdict | Version |
+|------|-----------|---------|---------|
+| 1. Incomplete port / maintenance | wafer-inc has meaningful gaps for LLM validation | **FALSE for this use case** — calendar math works through 2030+; but see day-of-week caveat | Inform 0.2.0 README |
+| 2. GVL thread blocking | 500µs holds noticeably degrade concurrent Ruby | **PLAUSIBLE** — 505µs measured; Magnus 0.9 has no high-level GVL release API | Benchmark at 0.2.1; fix in 0.3.0 if needed |
```
cpb: **suggestion:** do we have a ticket for setting up benchmarks? I'd love some benchmark-ips built in with automatic reporting the latest benchmark in the readme and notable changes in the changelog

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
```diff
@@ -0,0 +1,405 @@
...
+| 3. Panic boundary crash | Panics inside duckling crash the Ruby process | **ALREADY MITIGATED** by duckling's own `catch_unwind` in release builds | Code practice for 0.2.0; not architectural |
+| 4. GC pressure | Ephemeral Ruby object allocation spikes GC | **UNKNOWN** — plausible at scale; not measurable without profiling at production load | Benchmark at 0.2.1 |
```
cpb: **note:** this gc pressure should be part of the benchmark issue: https://github.com/cpb/duckling/pull/3/changes#r3506832233

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/test-coverage/ruby-test-design.md
cpb: Filed as issue #34 (labeled `test-first`) — checked main and confirmed none of this doc's 7 designed test classes (`TestDucklingParseTimeBasic`, `TestDucklingParseTimeWeekdays`, etc.) were ever implemented; PR #2 shipped only the narrow 7-class hill-test suite needed to gate 0.2.0. This was already flagged as follow-on work in 00-pr2-roadmap.md but never turned into a tracked issue until now.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/test-coverage/pyduckling-reference.md
cpb: Both addressed:

- Replaced all three references to `pr_context.md` (this file and ruby-test-design.md) with links to GitHub issue #1, which carries the same acceptance-criteria text and is durable (56dff51).
- Filed the cross-referencing suggestion as issue #35, labeled `backlog` so it doesn't get swept up by an automated babysitting loop — it's exploratory research (audit wafer-inc-duckling's corpus against pyduckling's and the original Haskell duckling's test suites), not a scoped implementation task.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Impact analysis
+
+| Scenario | GVL hold | Impact |
+|----------|----------|--------|
+| Single Puma thread, LLM output validation | ~500 µs | Negligible |
+| 8 Puma threads, all parsing simultaneously | ~500 µs each, serialized | Up to 3.5ms worst-case wait |
+| Short phrases ("tomorrow") | ~500 µs | Low — typical for native ext calls |
+| Long LLM-generated paragraphs | ~3 ms | Noticeable — 3x slower than a typical Ruby context-switch quantum (~1ms) |
+| Streaming LLM validation per-chunk | Depends on chunk size | Could block if chunks are large prose |
```
cpb: **polish:** how about with falcon? how about in sidekiq?

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
cpb: No existing ticket — filed as issue #36: a `benchmark-ips` suite for `Duckling.parse`, with results kept current in the README and notable changes surfaced in the CHANGELOG.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
cpb: Folded in — issue #36's acceptance criteria explicitly cover both Risk 2 (GVL/timing) and Risk 4 (GC pressure/allocation) as one benchmark suite, not two separate tickets.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
cpb: Added in a46c40c — two new subsections in ffi-risks.md:

- **Falcon**: different failure mode than Puma, worth calling out explicitly. Falcon multiplexes Fibers onto a small number of OS threads via cooperative yielding (the `async` gem). A blocking, non-yielding native call like `Duckling.parse` doesn't yield the Fiber scheduler either, so it stalls the *entire reactor thread* for its duration — every other concurrent Fiber on that worker, not just one OS thread contending for the GVL among several. That's a meaningfully worse story than Puma's "N threads serialize" framing.
- **Sidekiq**: thread-based like Puma, so the same GVL-contention math applies directly, but the practical impact is lower — Sidekiq workloads are throughput-oriented (background jobs), not latency-sensitive web requests, so a few hundred µs to a few ms of added serialization per job is unlikely to matter except at very high job throughput or for latency-sensitive job types.

Both are reasoning from each server's concurrency model, not new measurement — updated issue #36 (the benchmark-ips suite) to include Falcon and Sidekiq scenarios so this gets validated empirically rather than staying theoretical.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
```diff
@@ -0,0 +1,448 @@
+# FFI Binding Risks — Hypotheses and Test Results
```
cpb: **polish:** this needs a toc

cpb: Added in 5167218 — a Table of Contents right after the intro, linking each Risk section (1-6), the two new Falcon/Sidekiq subsections under Risk 2, plus the Version Roadmap and Appendix at the end.

docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md
```diff
@@ -0,0 +1,448 @@
...
+thread of execution.
+
+Practical implication: on Falcon, the "8 threads serialized" row above doesn't apply the
+same way — there's no serialization queue in the usual sense, just one reactor thread
+stalled for the call's duration, which delays *every* concurrent request on that worker,
+not up to N of them. For short inputs this is likely still negligible in absolute terms
+(~500µs), but for long prose (~3ms) it's a larger fraction of a Falcon worker's available
+scheduling slice than the equivalent Puma-thread contention, precisely because Falcon
+workers are typically provisioned assuming handlers don't block synchronously for
+multiple milliseconds at a time.
```
cpb: **suggestion:** this needs an issue to explore, test driven (ie: arranging calls in multiple fibers etc)

cpb: Filed as issue #38 (labeled `test-first`), specifically as you described: arrange multiple concurrent Fibers via the `async` gem (one calling `Duckling.parse`, others doing observable timed work), measure whether the others get delayed by the parse call's duration, and write the actual result back into ffi-risks.md's Falcon subsection — replacing the current "reasoning from the concurrency model" framing with a measured one.

This overlapped with a bullet already in issue #36 (the benchmark-ips suite) — trimmed that bullet and pointed it at #38 instead, so the Fiber-blocking claim has one dedicated, test-first home rather than being tracked in two places.

cpb:
**Note on a deleted review comment:** a review comment was left on `plans/README.md` ("This seems fully accomplished with 0.2.0 -- can you separate out the stuff which no longer applies into a stale subpath, and leave only the details that remain as a plan for a 0.2.x?"), and I accidentally deleted it (and the review it was part of) via a bad `DELETE` API call while cleaning up an unrelated error. Recording it here since I can't restore it or reply inline anymore.

**What I did in response**, already committed and pushed in 8e0375d:

- Moved the four fully-executed implementation plans (00-pr2-roadmap, 01-native-extension-setup, 02-ruby-api-design, 03-test-suite-and-ci) into `docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/plans/stale/`, each with a "STALE — fully executed" banner and a `stale/README.md` explaining the archive. Verified all four match `main@03a69e1` exactly (v0.2.0 shipped, PR #2 merged).
- Rewrote `plans/README.md` as the live plan: what's actually left, organized by theme (environment/tooling alignment, Rakefile workflow, API design exploration, test coverage, performance) and linked to the GitHub issues already tracking each item, instead of restating prose that would go stale again.
- While archiving, found three genuinely-open items that weren't tracked as issues yet — `reference_time:` Ruby Time object support, the remaining 13 `DimensionValue` variants, and upstreaming serde attributes to wafer-inc/duckling — and filed them as #45, #46, #47 so nothing got silently dropped in the move.
- Updated the top-level `README.md`'s reading order, table of contents, and "Open Questions" section to match.

Issue #27, opened 2026-07-01
Fill in bin/claude-code-web-setup with just-in-time dependency install


cpb:
## Goal

Replace the no-op `bin/claude-code-web-setup` skeleton with a real setup hook that installs whatever the project needs (gems now; Rust toolchain / native extension compile once it lands) just-in-time in the Claude Code Web remote sandbox.

## Background

`bin/claude-code-web-setup` runs before Edit/Write tool calls in remote Claude Code sessions (`CLAUDE_CODE_REMOTE=true`) but is currently a pure no-op (`exit 0`). Per the referenced environment observation ([fizzbuzz_app's doc/observation/2026-06-07-claude-code-web-environment.md](https://github.com/cpb/fizzbuzz_app/blob/main/doc/observation/2026-06-07-claude-code-web-environment.md)), the Claude Code Web sandbox already ships Ruby 3.3.6 (via rbenv) and Rust 1.94.1 (rustc/cargo) pre-installed — so this hook mainly needs to run `bundle install` and, once the native extension (`ext/duckling/`) lands, confirm the Rust extension compiles, rather than installing language runtimes from scratch.

The skeleton's own comment recommends using `tmp/claude-web-receipts/` for receipt-file caching so setup doesn't re-run on every tool call, pointing to fizzbuzz_app's `bin/claude-code-web-setup` as a reference implementation.

Raised during PR #3 review (research/planning PR for issue #1). Not blocking PR #3 (docs-only).

## Acceptance criteria

- [ ] `bin/claude-code-web-setup` runs `bundle install` (skipping if already satisfied, via receipt-file caching in `tmp/claude-web-receipts/`)
- [ ] Once the native extension lands, the hook also verifies/builds it (e.g. `rake compile`) so `require "duckling/duckling"` works without the agent needing to run setup manually
- [ ] A fresh remote Claude Code Web session on this repo can edit/run project code without a human running manual setup commands first

## Out of scope

- Installing Ruby or Rust themselves (already present in the Claude Code Web image per the referenced observation doc)

## Notes

Reference implementation: fizzbuzz_app's `bin/claude-code-web-setup` (receipt-file caching pattern).

cpb:
Closed by #42 (merged).

PR #42, opened 2026-07-01
Fill in bin/claude-code-web-setup with just-in-time dependency install

cpb:
## Summary
- Replaces the no-op `bin/claude-code-web-setup` skeleton with a receipt-cached setup hook, closing #27
- Before each Edit/Write in a Claude Code Web session, installs gems (`bundle install`), compiles the native extension (`bundle exec rake compile`), and provisions `hk` (installs the binary + `hk install` if missing) — each step cached under `tmp/claude-web-receipts/` so it's a no-op after the first call per session
- Extension-compile receipt hashes `ext/duckling`'s source files directly (not just `Cargo.lock`, which doesn't change from local `.rs` edits since the root crate isn't a checksummed dependency in it)
- `hk`'s version is parsed out of `hk.pkl`'s pinned `amends` line rather than hardcoded a second time, so it can't drift from the actual lint config
- Updates the stale AGENTS.md description of this hook
- Adds `.claude/settings.json` registering the `cpb` marketplace and enabling the `cpb@cpb` plugin at the project level (mirroring `fizzbuzz_app`)
- Inlines the three PreToolUse/PostToolUse hooks directly in `.claude/settings.json`, pointing straight at the project's own `bin/check-worktree` / `bin/claude-code-web-setup` / `bin/lint`. Registering the marketplace alone wasn't sufficient — a live `$CLAUDE_CODE_REMOTE=true` session on this branch still didn't fire the hooks, most likely because the marketplace-trust approval Claude Code prompts for on a newly-declared repo marketplace is an interactive flow that isn't answerable in a headless/remote session. Filed cpb/claude-plugins#25 so `/cpb:init` does this by default for future projects.
- Extracts `install_gems`/`compile_extension` into a shared `bin/claude-web-deps.sh`, sourced by both `bin/claude-code-web-setup` and `bin/test`'s new no-arg path (`bundle exec rake`, JIT-installing gems first in remote sessions) — a bare `bin/test`/`bundle exec rake test` via Bash doesn't trigger the Edit/Write-gated hook, so it failed cold in the same live-session test
- Makes `install_hk`/`install_hk_hooks` best-effort: a live-session test hit a `curl: (22) 403` provisioning `hk` (likely the sandbox's network egress allowlist not permitting the GitHub release-asset redirect target), which under `set -eu` surfaced as a hook error blocking the edit entirely and would have cascaded into `bin/lint` failing on every subsequent edit — failures here now warn on stderr instead

## Test plan
- [x] `sh -n bin/claude-code-web-setup`, `bash -n bin/test`, `bash -n bin/claude-web-deps.sh` — syntax checks
- [x] Cold-cache run → confirmed `bundle install` + `bundle exec rake compile` + `hk install` all ran, artifact + receipts produced
- [x] Warm-cache re-run → confirmed full no-op (~60ms, no install output)
- [x] Content change to `ext/duckling/src/lib.rs` → confirmed only the extension receipt invalidated and recompiled
- [x] `bundle exec rake` and `bin/test` (no-arg, local and simulated-remote-cold) pass
- [x] Isolated test of `install_hk` with `hk`/`brew` hidden and `curl` faked to fail (403) → confirmed no receipt written and the script continues past the failure instead of aborting
- [x] Live remote-session test surfaced both the marketplace-trust gap and the two gaps above; this PR's later commits address them
- [ ] Re-verify in a fresh remote Claude Code Web session: inline hooks fire, `bin/test`/`bundle exec rake test` succeeds cold via Bash, and an `hk` provisioning failure no longer blocks edits

https://claude.ai/code/session_018c4xr2vKDVVDQMuem99AW4


cpb:
Ran the test suite in a Claude Code Web session against this branch. `bundle exec rake` initially failed with `bundler: command not found: rake` even after `bin/setup` completed successfully — the session's default `ruby`/`gem`/`bundle` resolve to `/opt/ruby-3.3.6`, but the gems actually install under the rbenv-managed `3.3.6` install, and plain `bundle exec` doesn't pick up the right `PATH`/version there.

Fixed by exporting these before running any `bundle exec` command:

```bash
export PATH="/opt/rbenv/shims:$PATH"
export RBENV_VERSION=3.3.6
```

With those set, `bundle exec rake` runs clean: StandardRB lint passes, the Rust extension compiles, and Minitest reports 16 runs / 52 assertions / 0 failures / 0 errors (3 skips).

Flagging in case `bin/claude-code-web-setup` (or session setup docs) should export these too, so future web sessions don't hit the same `command not found: rake` failure.

---

PR #44, opened 2026-07-02
Implement bin/claude-code-web-setup: JIT gem/extension/hk install for Web sessions

cpb:
Replaces the no-op skeleton with a receipt-cached setup hook, closing
issue #27. A fresh Claude Code Web session previously had no gems, no
compiled native extension, and possibly no hk binary, so both the
project's Ruby code and the bin/lint PostToolUse hook would fail.

Modeled on fizzbuzz_app's reference implementation, with two changes:
the extension-compile receipt hashes ext/duckling's source files
directly (Cargo.lock alone doesn't change from local .rs edits, since
the root crate isn't a checksummed dependency in it), and the hk
version is parsed out of hk.pkl's pinned amends line instead of being
hardcoded a second time, so it can't drift from the actual lint config.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_018c4xr2vKDVVDQMuem99AW4


AGENTS.md
```diff
@@ -30,7 +30,7 @@ real implementation lands (see "Keeping this file current").
...
 | `Rakefile` | `task default: %i[standard compile test]` — runs StandardRB lint, compiles the Rust extension, then Minitest. |
 | `.standard.yml` | StandardRB config — see its `ruby_version:` field for the Ruby version StandardRB targets. StandardRB wraps RuboCop internally; there is no separate `.rubocop.yml`. |
-| `hk.pkl` | `hk` config (StandardRB + rustfmt + clippy via `hk`'s builtin steps) — single source of truth for local lint/format enforcement. `bin/lint` runs `hk fix` against it; not used by CI (CI runs the underlying tools directly, see below). |
+| `hk.pkl` | `hk` config (StandardRB + rustfmt + clippy via `hk`'s builtin steps), kept for anyone running `hk` manually/in CI environments with GitHub access. `bin/lint` no longer shells out to it (see below) — it runs the same underlying tools directly, since `hk`'s Pkl config needs to fetch its schema package from a GitHub release on every invocation. Not used by CI either (CI runs the underlying tools directly, see below). |
```
cpb, line 33: **question:** what other indications of hk are in the repo? do we have the pre-commit hooks setup? Would that impact the remote claude environments if so? Is the dependency installed in the Brewfile?

cpb:
Closing as superseded by #42, which is the tracked implementation of issue #27 (more extensive test plan, live-session testing/fixes already applied).

Your `bin/lint` rewrite here was the right call, though — dropping the `hk fix` GitHub-schema-fetch dependency fixes the exact curl-403-in-sandbox problem #42 could only work around with a best-effort/warn-on-failure hack. Ported it over in cpb/duckling@7b4be64 on #42's branch. Thanks for surfacing it.

cpb:
Good catch — answered in #52 (merged): `hk` had no automated local install path (missing from the Brewfile), and `bin/claude-code-web-setup` was still installing it + running `hk install` in every remote session even though `bin/lint` no longer needs it (per the fix from this PR, ported into #42). Resolved by scoping `hk` to local dev only: added to `Brewfile`, `bin/setup` now runs `hk install` to wire up the git pre-commit hook automatically, and all remote-session `hk` provisioning was removed from `bin/claude-code-web-setup`.

PR #50, opened 2026-07-02
Implement thread-per-call GVL release to unblock async reactor Fibers (#64)

### LINKED ISSUE #38: Test-drive the Falcon Fiber-blocking claim for Duckling.parse
## Goal

Write a test that empirically verifies (or falsifies) the theoretical claim in [`docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/ffi-risks.md`](https://github.com/cpb/duckling/wiki/research-ffi-risks) (Risk 2, Falcon subsection): that a blocking `Duckling.parse` call stalls every other Fiber sharing the same Falcon/`async` reactor thread for the duration of the call, not just contending for the GVL among OS threads the way Puma's thread model does.

## Background

That subsection was added reasoning from how Falcon's concurrency model works (Fibers cooperatively multiplexed onto few OS threads via the `async` gem; a call that doesn't yield the GVL doesn't yield the Fiber scheduler either) — it has not been empirically tested. A PR #3 review comment asked for this to be explored test-first, specifically by "arranging calls in multiple fibers."

This is distinct from the general `benchmark-ips` reporting suite in issue #36 — that issue produces headline timing/allocation numbers for the README; this issue is about proving or disproving a specific concurrency-model claim with a targeted test, which is a different kind of artifact (a pass/fail or measured-delta test, not a benchmark report).

## Acceptance criteria

- [ ] A test (using the `async` gem directly, without necessarily standing up a full Falcon HTTP server) arranges at least 2-3 concurrent Fibers/tasks on one reactor: one calls `Duckling.parse` with a representative input (e.g. the "long LLM-generated paragraph" case, ~3ms), the others perform some observable, timed work (e.g. incrementing a counter on a tight loop, or scheduled to run at a known interval)
- [ ] The test measures whether the other Fibers' scheduled work is delayed by approximately the `Duckling.parse` call's duration (confirming the "de-fiberizes the reactor" claim) or continues running unaffected (falsifying it)
- [ ] Results are written back into `ffi-risks.md`'s Falcon subsection, replacing "reasoning from the concurrency model" with an actual measured result
- [ ] If the claim is confirmed, note whether Falcon offers any built-in mitigation (e.g. running the blocking call in a separate thread pool via `Async::Task` or similar) worth documenting as a workaround for 0.2.0 users on Falcon

## Out of scope

- General benchmark-ips timing/allocation reporting for the README/CHANGELOG (tracked in #36 — that issue's Falcon bullet is trimmed to point here instead, to avoid duplicate tracking)
- Fixing the blocking behavior itself if confirmed (that's a 0.3.0-scale GVL-release concern, same as Risk 2's existing fix path)

## Notes

Labeled `test-first`: write the concurrent-Fiber test as the hill before drawing conclusions, since the whole point is to replace a theoretical claim with a measured one.

### LINKED ISSUE #64: Implement thread-per-call GVL release to unblock async reactor Fibers in Duckling.parse
## Goal

Make `Duckling.parse` stop stalling sibling `Fiber`s on an `Async::Reactor` (Falcon) during its native call — supersedes the implementation scope of #57 with the concrete, empirically-grounded plan below (research for #57 has moved to [the wiki](https://github.com/cpb/duckling/wiki/Home); #57 itself stays open as the research-tracking issue).

## Background

`Duckling.parse` is a synchronous Rust FFI call that holds Ruby's GVL for its entire duration. Inside an `Async::Reactor`, which cooperatively schedules `Fiber`s on a single OS thread, this stalls every sibling Fiber for the call's duration — `test/falcon_fiber_blocking_test.rb` (from #38) proves this empirically and is the failing hill test this issue exists to turn green.

An empirical spike (see [research-fiber-scheduler-mechanism-spike](https://github.com/cpb/duckling/wiki/research-fiber-scheduler-mechanism-spike) and its [results](https://github.com/cpb/duckling/wiki/research-fiber-scheduler-mechanism-spike-results)) found the headline result driving this whole issue: **releasing the GVL alone is not enough**, even on Ruby 3.4.5. `rb_thread_call_without_gvl` never sets the `RB_NOGVL_OFFLOAD_SAFE` flag that Ruby 3.4's `Fiber::Scheduler#blocking_operation_wait` auto-offload path requires, so a bare GVL release doesn't hand control back to the reactor. The fix needs **both**:

1. A raw GVL release around the native call (`rb_thread_call_without_gvl` — Magnus 0.8.2 has no safe wrapper for this; see [research-magnus-rb-sys-gvl-release](https://github.com/cpb/duckling/wiki/research-magnus-rb-sys-gvl-release)), **and**
2. A genuine background `Thread` spawned at the Ruby level (`Thread.new { ... }.value`), so the calling Fiber can yield to the reactor via `Thread#value`'s `block`/`unblock` scheduler hooks (present since Ruby 3.0, unlike `blocking_operation_wait`).

The combination passed 11/11 runs across Ruby 3.3.6 and 3.4.5 in the spike; neither piece alone did. Dispatch is **thread-per-call** (`Thread.new` per invocation), not a persistent worker pool — see [research-concurrency-alternatives-comparison](https://github.com/cpb/duckling/wiki/research-concurrency-alternatives-comparison) — since issue #57 has no throughput-optimization goal, and thread-per-call gets panic/exception propagation for free via `Thread#value`'s re-raise semantics.

The wrapped `duckling` crate is confirmed safe to call concurrently ([research-duckling-crate-thread-safety](https://github.com/cpb/duckling/wiki/research-duckling-crate-thread-safety)): no unsynchronized global state, every cache is `Mutex`/`OnceLock`-guarded and held only briefly, every public type is `Send + Sync`. The one open precondition is panic safety: `duckling::parse`'s own internal `catch_unwind` is compiled out entirely under `#[cfg(not(debug_assertions))]` (i.e. absent in this repo's own `dev`-profile local default), so the wrapper's own guard must be unconditional, not release-profile-only.

## A new seam: `Duckling::Native.parse`

As part of this issue, split the native singleton method out from `Duckling.parse` into a new `Duckling::Native` module, so `Duckling.parse` becomes a thin Ruby-level wrapper:

```ruby
module Duckling
  def self.parse(*args, **kwargs)
    Thread.new { Native.parse(*args, **kwargs) }.value
  end
end
```

This is a small, independent module split scoped only to this issue — just moving the existing `RArray`-returning native method under `Duckling::Native`, no return-type changes. (Note: PR #53, open and unmerged, separately introduces a similar `Duckling::Native` split, but bundles it with an unrelated serde_magnus/`Data`-object return-type change for issue #32 — this issue does **not** depend on #53 merging; treat the split here as freshly, minimally scoped.)

The reason this seam matters: it lets a benchmark call `Duckling::Native.parse` directly (no thread, the "without" baseline) versus `Duckling.parse` (thread-per-call dispatch, the "with" case) to directly measure the cost of the fix — see the benchmark acceptance criterion below.

## Concrete implementation shape

In `ext/duckling/src/lib.rs`:

1. Add a `ParsePayload` struct carrying owned `text`/`locale`/`dims`/`context`/`options` in, and `Option<Result<Vec<Entity>, String>>` out — deliberately only plain Rust data, zero `magnus::Value`/`magnus::Error` in either direction (this repo's established rule: never stash a bare `magnus::Value`, or anything wrapping one, across a Magnus call boundary — a past incident here caused a real GC-safety segfault).
2. Add an `unsafe extern "C" fn parse_without_gvl(*mut c_void) -> *mut c_void` callback that runs `duckling_parse(...)` inside `std::panic::catch_unwind` (the `std` version directly, not `magnus::rb_sys::catch_unwind` — keeps the payload's "plain Rust data only" invariant simple and mechanically checkable).
3. Rename/expose the existing `parse` function as `Duckling::Native.parse`; its body boxes the payload (`Box::into_raw`), calls `rb_sys::rb_thread_call_without_gvl(Some(parse_without_gvl), payload_ptr, None, ptr::null_mut())`, then `Box::from_raw`s it back. No `Cargo.toml` change needed — `rb-sys` exposes this FFI binding unconditionally already.
4. On the `Err(message)` branch, construct `magnus::Error::new(ruby.exception_fatal(), ...)` only *after* the box is reclaimed (GVL confirmed held again) — never inside `parse_without_gvl`.
5. In `lib/duckling.rb`, define `Duckling.parse` to wrap `Duckling::Native.parse` in `Thread.new { ... }.value`, per the seam above.
6. Decide where argument conversion (`scan_args`/`get_kwargs`, `parse_locale`/`parse_dims`/`build_context`) runs relative to the Thread-spawn — converting on the calling thread first, before spawning, is the safer default (confines the spawned thread's Ruby-object surface to nothing) but needs verifying against the hill test, not assumed.

Full illustrative sketch (not to be copied verbatim, but mapped onto the real file): [research-magnus-rb-sys-gvl-release-implementation-sketch](https://github.com/cpb/duckling/wiki/research-magnus-rb-sys-gvl-release-implementation-sketch).

## Acceptance criteria

- [ ] `test/falcon_fiber_blocking_test.rb` passes (max ticker gap stays under the test's `0.011s` threshold during a concurrent `Duckling.parse` call)
- [ ] `Duckling::Native.parse` exists as the raw native entrypoint (no Thread spawn, no GVL release considerations visible at the call site); `Duckling.parse` wraps it in `Thread.new { ... }.value`, thread-per-call, no persistent worker pool
- [ ] GVL is released around the native `duckling_parse` call via `rb_thread_call_without_gvl`, guarded by an unconditional `std::panic::catch_unwind` (not compiled out in `dev` profile — must not rely on the wrapped crate's own release-profile-only guard)
- [ ] Existing tests (`test/duckling_test.rb`, `test/duckling_comma_list_test.rb`, and any others added since) stay green — return values unchanged
- [ ] `bundle exec rake` (full suite: `standard` + `compile` + `test`) passes
- [ ] **Benchmark with side-by-side charts comparing with-thread vs. without-thread dispatch overhead.** This repo already has a `benchmark-ips`/GC/concurrency harness (`benchmark/parse_benchmark.rb`, driven via `bin/benchmark record-pr`, which writes JSON history under `docs/benchmarks/<environment>/<version>.json` and regenerates `docs/benchmarks/README.md` including `mermaid xychart-beta` bar charts — see that README for the current shape). Extend it with a dispatch-mode dimension: run the existing `CORPUS` scenarios (`short`/`medium`/`long`/`no_match`/`empty`/`camping_trip_email`) against **both** `Duckling::Native.parse` (no thread — the baseline) and `Duckling.parse` (thread-per-call — the fix), and add a comparison chart analogous to the existing per-environment ips chart (grouped bars: no-thread vs. with-thread, per scenario), plus the measured overhead percentage per scenario. Prior research measured thread-spawn cost at ~70µs/call locally — expect this to be a rounding error against `long` (~3.8ms) and `camping_trip_email` (~790ms) but a real, visible 2-4x latency multiplier on the fastest scenarios (`empty` ~24µs, `no_match` ~213µs) — the chart should make this tradeoff visible, not just pass/fail.
- [ ] Panic-safety: a panic inside `duckling::parse` (dev-profile build, where the wrapped crate's own `catch_unwind` doesn't exist) is still caught by the wrapper's own guard and surfaces as an ordinary raised Ruby exception via `Thread#value`'s re-raise semantics, not an aborted process

## Out of scope

- General parse-throughput optimization beyond removing the reactor-blocking behavior (a worker pool, bounded concurrency, etc. — see [research-concurrency-alternatives-comparison-worker-pool](https://github.com/cpb/duckling/wiki/research-concurrency-alternatives-comparison-worker-pool) for why thread-per-call was chosen instead)
- A conditional/opt-out thread-spawn for non-reactor callers (e.g. `Fiber.current_scheduler`-based detection to skip the Thread when no reactor is present) — the detection signal was analyzed as sound but unexercised in any test; tracked separately as a roadmap follow-up, not part of this issue
- A `ubf` (unblocking function) cancellation hook for `Thread#raise`/`Thread#kill` against an in-flight parse — deferred, tracked separately
- Bumping the Ruby version floor (currently `>= 3.2.0`) — Ruby 3.4's `Fiber::Scheduler#blocking_operation_wait` would not help this specific fix anyway unless the native code separately called the lower-level `rb_nogvl` with `RB_NOGVL_OFFLOAD_SAFE` set, which is a distinct, later decision

## Source research

Full research (5 topics, empirical spike, thread-safety/panic-safety audit of the wrapped crate, and test/CI-mechanics confirmation) lives on the wiki — start at [Home](https://github.com/cpb/duckling/wiki/Home) or the topic index at [research-async-reactor-blocking](https://github.com/cpb/duckling/wiki/research-async-reactor-blocking).

cpb:
## Summary

Closes #38, Closes #64.

Stacked on this branch's own hill commits (originally the failing-test PR for
issue #38) — implements the fix and merges it in.

- Splits the native entrypoint out to `Duckling::Native.parse` (no thread spawn, no GVL-release considerations at the call site). `Duckling.parse` (`lib/duckling.rb`) is a thin Ruby-level wrapper that dispatches through `Thread.new { Native.parse(...) }.value` **only when a `Fiber.scheduler` is installed on the calling thread** — a plain thread pool (Puma/Sidekiq-style, no reactor to yield to) already gets its concurrency from `Native.parse`'s own GVL release, so it skips the thread spawn entirely and calls `Native.parse` directly.
- `Duckling::Native.parse` releases the GVL around `duckling::parse` via the raw `rb_thread_call_without_gvl` FFI (Magnus 0.8.2 has no safe wrapper for this). Only plain owned Rust data (`ParsePayload`, stack-local, no heap allocation) crosses the off-GVL boundary — no `magnus::Value`/`magnus::Error` in either direction, per this repo's established GC-safety rule.
- The off-GVL callback wraps the call in `std::panic::catch_unwind` unconditionally, since the wrapped `duckling` crate's own panic guard is compiled out under `debug_assertions` (this repo's own `dev`-profile local default). A caught panic surfaces to the caller as a rescuable `RuntimeError` with the original message preserved (not magnus's own `Error::from_panic` convention of the unrescuable `fatal`).
- A bare GVL release alone is *not* sufficient to unblock an `Async::Reactor` (confirmed by the empirical spike on [the wiki](https://github.com/cpb/duckling/wiki/research-fiber-scheduler-mechanism-spike) — Ruby 3.4's `Fiber::Scheduler#blocking_operation_wait` needs a flag `rb_thread_call_without_gvl` never sets) — the `Thread.new` piece is what actually lets the calling Fiber yield via `Thread#value`'s scheduler hooks. The spawned thread also disables `report_on_exception` from its first line, so a rescued error doesn't also print a thread-termination backtrace to stderr.
- Extends the benchmark suite with a dispatch-mode dimension: `Duckling::Native.parse` (no thread) vs `Duckling.parse` (thread-per-call, measured under an active Fiber scheduler via a `Sync` block — see "Benchmark data" below for why that matters), with a grouped-bar chart and per-scenario overhead percentage in the generated `docs/benchmarks/README.md`. Gracefully omitted for older recorded entries that predate this schema.
- Updated `AGENTS.md`'s `lib/duckling.rb` directory-layout entry and "Rust/Magnus wiring" section to document the `Fiber.scheduler` gating, the rescuable-panic behavior, and (per review) that this gem's `require_relative`-based loading has no interaction with a consuming app's Zeitwerk setup. Added a "Code comment conventions" section documenting that comments should explain durable rationale, not narrate which issue/PR added them.
- Dropped `test/duckling_test.rb`'s hardcoded `Duckling::VERSION` equality test (per review) — needed a manual bump every release with no regression-catching value.

### Additional hardening from review

An earlier review round (findings posted inline, most-severe first) caught:
- Unconditional thread-per-call taxing every synchronous caller — fixed via the `Fiber.scheduler` gating above.
- The stderr backtrace noise on rescued errors — fixed via `report_on_exception = false`.
- The unrescuable `fatal` panic class — fixed via a rescuable `RuntimeError`.
- Along the way, found and fixed a real bug independent of the above: `panic_message`'s downcast took `&panic_payload` where `panic_payload: Box<dyn Any + Send>` — since `Box<dyn Any + Send>` itself implements `Any`, that reference coerced to treat the *Box* as the payload rather than dereferencing into it, so the downcast always missed and every panic message was silently discarded. Fixed to `&*panic_payload`.

## Benchmark data

Also found and fixed a bug in `run_ips`'s own dispatch-mode measurement:
`x.report(name) { ... }` inside `Benchmark.ips do |x| ... end` only
*registers* a block — the timed execution (`job.run`) happens after that
surrounding block returns. Wrapping just the registration loop in `Sync do
... end` therefore measured nothing (by execution time, `Fiber.scheduler`
was already back to `nil`), producing backwards results (dispatch reading
as *faster* than native on some scenarios). Fixed by wrapping the entire
`Benchmark.ips` call — registration and `job.run` — in one `Sync` block, so
the dispatch scenarios actually take the thread-spawn path they're meant to
measure. Re-verified: `empty` (the fastest, most overhead-sensitive
scenario) now reads ~11x slower dispatched vs native, consistent in
direction and rough magnitude with `github-actions`/`claude-code-web`'s
existing recorded data; `long`/`camping_trip_email` show negligible
overhead, as expected.

Re-recorded `docs/benchmarks/local/0.2.1-rc1.json` with the fixed
methodology and the `Fiber.scheduler`-gating code change together.

**Known-stale data, not fixed in this PR:** `docs/benchmarks/github-actions/0.2.1-rc1.json`
and `docs/benchmarks/claude-code-web/0.2.1-rc1.json` were recorded *before*
the `Fiber.scheduler`-gating fix landed, when `Duckling.parse` unconditionally
spawned a Thread on every call. Their dispatch-overhead (ips vs native_ips)
numbers are still directionally valid (that code path always paid the
thread-spawn tax back then, same effect as today's `Sync`-wrapped
measurement) — but their **concurrency/throughput** numbers are now stale:
`efficiency_pct` reads 20.4% (github-actions) / 23.7% (claude-code-web) vs
local's fresh 62.6%, purely because the plain-thread-pool concurrency
benchmark no longer pays a per-call thread-spawn tax after the gating fix,
and only `local`'s data reflects that yet. In the generated README this
currently reads as an environment/hardware difference when it's actually a
code-version difference within the data itself.
- `github-actions` can't be refreshed pre-merge: `benchmark-branch.yml`
  (this PR's own generalized, `workflow_dispatch`-only replacement for the
  old push-triggered scaffolding) isn't dispatchable against this branch
  until it exists on `main` — confirmed via `gh workflow run` (404). Refresh
  via `gh workflow run benchmark-branch.yml --ref main` (or the normal
  `benchmark.yml` dispatch) once merged.
- `claude-code-web` can only be recorded from an actual Claude Code Web
  remote session (`CLAUDE_CODE_REMOTE=true`) — needs a follow-up
  `bundle exec rake benchmark:record_pr` run from such a session.

## Test plan
- [x] `test/falcon_fiber_blocking_test.rb` (the failing test this hill introduced) now passes
- [x] `test/thread_pool_dispatch_test.rb` — plain thread-pool callers (no `Fiber.scheduler`) spawn zero per-call `Thread`s
- [x] `test/parse_error_stderr_test.rb` — rescued errors write nothing to stderr
- [x] `test/native_panic_test.rb` — a native panic surfaces as a rescuable `StandardError` with its message preserved, and the process survives
- [x] `bundle exec rake` (standard + compile + test) green
- [x] `cargo clippy -- -D warnings` and `cargo fmt --check` clean
- [x] Ran the real `bundle exec rake benchmark:record` (not just fixture-based unit tests) to confirm the corrected dispatch-mode ips comparison actually executes, produces sane (non-backwards) numbers, and is recorded to `docs/benchmarks/local/`
- [x] New unit tests for `DucklingBenchmark::Report.build_dispatch_section`, including a legacy-entry-without-native-data case (backward compat)

<details>
<summary>Test first (original hill PR description)</summary>

## Hill: failing tests for issue #38

This draft PR contains only failing tests — one commit of stubs (adding the `async`
dev dependency) and one commit of the test itself. CI should confirm the failure below.
**Do not merge — implementation happens on this branch.**

### falcon-fiber-blocking

```
Expected every ticker Fiber tick inside the Async::Reactor to stay within 0.011s of the
requested 0.001s interval (i.e. Duckling.parse does NOT stall sibling Fibers), but the
largest observed gap was 0.173s, while the measured Duckling.parse call itself took
0.1726s. A max gap approximately equal to the parse duration is exactly the signature
predicted by the blocking-FFI-call risk claim: Duckling.parse holds the GVL for its
entire native execution and never yields to the async reactor, so the ticker Fiber (and
every other Fiber on this reactor thread) is frozen for the duration of the parse call.
Expected 0.17297000001417473 to be < 0.011.
```

This is a real, empirically measured result (reproduced multiple times locally, gap ≈
parse duration each run), not a placeholder — the test's assertion encodes the
*non-blocking* null hypothesis on purpose, so a failure here is itself the empirical
confirmation of the Falcon Fiber-blocking claim in the
[`research-ffi-risks`](https://github.com/cpb/duckling/wiki/research-ffi-risks) wiki
page's risk analysis.

## Notes for the implementation session

- The test currently emits `Async::Task#sleep is deprecated, use Kernel#sleep instead`
  warnings from the installed `async` gem version — cosmetic, does not affect the result,
  but worth a quick look during implementation.

## Reviewer instructions

Review the expected failure message above. If it correctly specifies the intended
behavior per the issue, **remove the `hill-ready` label** from this PR — that signals
the implementation session to begin. Run `/continue-pr 38` (this PR) to start it.

</details>


cpb:
## Hill ready for review

CI is running — wait for it to confirm the failure is an **assertion failure** (not a load error).

**To proceed:**
1. Review the expected failure message in the PR description.
2. Open `/continue-pr 50` — that session will inspect CI and ask whether the hill is correctly specified.

cpb (commented):
Adversarial review — 8 findings, most-severe first, posted inline. Verification notes: the stderr-noise finding was reproduced empirically (`Thread.new { raise ArgumentError }.value` prints a thread-termination backtrace before `Thread#value` re-raises); the `fatal` panic class was checked against magnus 0.8.2 source (`Error::from_panic` also uses `exception_fatal`, so that part is parity, not a regression); and `docs/benchmarks/github-actions/0.2.0.json` / `claude-code-web/0.2.0.json` were confirmed to still hold pre-#64 native-dispatch numbers. The unsafe Rust itself checks out: the payload is freed on every path, no `magnus::Value` crosses the off-GVL boundary, and with `ubf: None` the callback can't be skipped. Findings 1–3 (stderr noise, unconditional thread-per-call, cross-environment chart mixing) are the ones I'd block on.

lib/duckling.rb
```diff
@@ -2,3 +2,19 @@
...
+  # control back to an Async::Reactor — Ruby 3.4's Fiber::Scheduler
+  # #blocking_operation_wait auto-offload path requires a flag
+  # rb_thread_call_without_gvl never sets. Spawning a real background Thread
+  # lets the calling Fiber yield to the reactor via Thread#value's
+  # block/unblock scheduler hooks instead, which have been present since Ruby
+  # 3.0. See the wiki's research-fiber-scheduler-mechanism-spike for the
+  # empirical result driving this. Call Native.parse directly to skip the
+  # thread-spawn (e.g. for benchmarking the dispatch overhead itself).
+  def self.parse(...)
+    Thread.new { Native.parse(...) }.value
```
cpb: **[correctness, confirmed]** Every exception `Native.parse` raises now also prints a thread-termination backtrace to stderr before `Thread#value` re-raises it — `Thread#report_on_exception` defaults to `true`, and joining/`value`-ing the thread does not suppress the report. Reproduced:

```
$ ruby -e 'begin; Thread.new { raise ArgumentError, "bad locale" }.value; rescue ArgumentError => e; puts "rescued: #{e.message}"; end'
rescued: bad locale
# stderr:
#<Thread:0x… run> terminated with exception (report_on_exception is true):
-e:3:in `block in <main>': bad locale (ArgumentError)
```

So `Duckling.parse("x", locale: "zz")` — a documented, rescuable `ArgumentError` path — now spams stderr on every rescued call. Fix:

```ruby
def self.parse(...)
  t = Thread.new { Native.parse(...) }
  t.report_on_exception = false
  t.value
end
```

cpb: **[efficiency/altitude]** Unconditional thread-per-call taxes every synchronous caller to serve the async-reactor minority — this PR's own recorded numbers show +53% (`short`), +123% (`no_match`), +965% (`empty`, 57,488 → 5,396 ips). Plain threaded callers (Puma, Sidekiq) already get concurrency from the GVL release alone and gain nothing from the extra thread. Since the `Native` split already exists, scheduler-aware dispatch is one line:

```ruby
def self.parse(...)
  Fiber.scheduler ? Thread.new { Native.parse(...) }.value : Native.parse(...)
end
```

Only callers actually on a Fiber scheduler pay the offload cost.

ext/duckling/src/lib.rs
```diff
@@ -54,7 +131,45 @@ fn parse(ruby: &Ruby, args: &[Value]) -> Result<RArray, Error> {
...
+    // documented step 4) before touching any Ruby Value again.
+    let boxed = Box::new(ParsePayload {
+        text,
+        locale,
+        dims,
+        context,
+        options,
+        result: None,
+    });
+    let payload_ptr = Box::into_raw(boxed) as *mut c_void;
```
cpb: **[simplification]** The `Box::new` → `Box::into_raw` → `Box::from_raw` round-trip (and the `Option<Result<…>>` + `.expect` on `result`) is avoidable: `rb_thread_call_without_gvl` is synchronous, so the payload never outlives this stack frame. A stack local does the same job with no heap allocation, no manual-free protocol to defend in comments, and no "can never happen" `.expect`:

```rust
let mut payload = ParsePayload { text, locale, dims, context, options, result: None };
unsafe {
    rb_sys::rb_thread_call_without_gvl(
        Some(parse_without_gvl),
        &mut payload as *mut ParsePayload as *mut c_void,
        None,
        std::ptr::null_mut(),
    );
}
let entities = match payload.result.take() { … };
```

(And if the callback returned the `Result` through its currently-discarded `*mut c_void` return value, the `Option` wrapper could go too.)

ext/duckling/src/lib.rs
```diff
@@ -54,7 +131,45 @@ fn parse(ruby: &Ruby, args: &[Value]) -> Result<RArray, Error> {
...
+    let boxed = unsafe { Box::from_raw(payload_ptr as *mut ParsePayload) };
+    let entities = match boxed
+        .result
+        .expect("parse_without_gvl always sets result before returning")
+    {
+        Ok(entities) => entities,
+        Err(message) => {
+            // Safe to construct/raise now: the GVL is confirmed held again.
+            return Err(Error::new(
+                ruby.exception_fatal(),
```
cpb: **[plausible]** Two notes on the panic path:

1. `exception_fatal` is parity with magnus 0.8.2's own `Error::from_panic` (verified against the published source), so this isn't a regression — but `fatal` is unrescuable (`rescue Exception` won't catch it), and it now has to propagate out of a background thread through `Thread#value`, which is untested territory. Since you're already catching the panic yourself, raising a rescuable `RuntimeError` would let an app survive a panicking input instead of dying — worth considering given the guard exists precisely because the dev profile compiles out the crate's own.

2. Cosmetic: `panic_message`'s fallback arm returns `"duckling::parse panicked"`, which this `format!` re-prefixes — a non-string panic payload yields `"duckling::parse panicked: duckling::parse panicked"`.

docs/benchmarks/README.md
```diff
@@ -71,13 +93,13 @@ xychart-beta
     y-axis "ips"
     bar "github-actions" [2820.1, 2314.0, 421.7, 7877.8, 73909.9]
     bar "claude-code-web" [2135.1, 1852.3, 366.0, 6362.9, 61108.3]
-    bar "local" [1474.4, 1448.1, 265.1, 4686.9, 41481.2]
+    bar "local" [1154.4, 1095.9, 279.2, 2477.0, 5396.2]
```
cpb: **[docs correctness]** This cross-environment chart now mixes measurement regimes: the `local` bars are post-#64 thread-per-call `ips`, while `github-actions` and `claude-code-web` still hold pre-#64 native-dispatch numbers (verified: their `0.2.0.json` files have no `native_ips`). The `empty` scenario reads as local being ~13x slower than GitHub Actions (5,396 vs 73,910) when almost the entire gap is the dispatch change, not hardware — and the README prose above explicitly invites cross-environment comparison as "meaningful." Options: re-record the other two environments before merging, plot `native_ips` here when present (it's the like-for-like series), or annotate the mixed regime.

docs/benchmarks/README.md
```diff
@@ -71,13 +93,13 @@ xychart-beta
...
 ```
 
 ```mermaid
 xychart-beta
     title "10-thread concurrency scaling efficiency (%) -- latest run per environment"
     x-axis [github-actions, claude-code-web, local]
     y-axis "efficiency %"
-    bar "efficiency_pct" [9.9, 9.4, 10.1]
+    bar "efficiency_pct" [9.9, 9.4, 38.4]
```
cpb: **[docs correctness]** Same regime-mixing as the throughput chart above: local's 38.4% efficiency was measured through thread-per-call dispatch (which slowed the single-thread baseline 1,643 → 1,048 ops/sec, inflating the scaling factor), while 9.9/9.4 are pre-#64 native numbers. The chart presents a methodology change as a 4x environment difference.

benchmark/parse_benchmark.rb
```diff
@@ -124,9 +131,19 @@ def self.run_concurrency
...
     scenarios = CORPUS.map do |s|
+      thread_stats = ips.fetch(s[:name].to_sym)
+      native_stats = ips.fetch(:"#{s[:name]}#{NATIVE_LABEL_SUFFIX}")
+      overhead_pct = ((thread_stats[:microseconds_per_call] - native_stats[:microseconds_per_call]) /
+        native_stats[:microseconds_per_call]) * 100
+
       {name: s[:name], input: s[:input]}
-        .merge(ips.fetch(s[:name].to_sym))
+        .merge(thread_stats)
         .merge(measure_gc(name: s[:name], text: s[:input]))
```
cpb, line 163: **[measurement validity]** `measure_gc` still dispatches through `Duckling.parse`, so the `objects/call` / `minor GC` columns now measure `Thread.new` churn, not parse allocation — the new JSON shows objects/call 28 → 35 and minor GC 1 → 62 purely from the wrapper. These columns render side-by-side with pre-#64 entries from other environments in `build_results_table`, reading as a 3–20x allocation regression. If the columns are meant to describe parse behavior, `measure_gc` should call `Duckling::Native.parse`; the same consideration applies to `run_concurrency`, whose single-thread baseline is now dominated by per-call spawn cost.

test/falcon_fiber_blocking_test.rb
```diff
@@ -0,0 +1,147 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "async"
+
+# Empirically tests the claim (from the FFI risk analysis) that a blocking
+# `Duckling.parse` call — because it's a synchronous Rust FFI call that does
```
cpb: **[stale docs]** This docstring (and the assertion's failure message at the bottom of the file) states categorically that `Duckling.parse` "does not release the GVL" and "holds the GVL for its entire native execution" — the opposite of what this same PR ships. It reads correctly as the historical hill claim, but a future reader (or a CI flake printing the failure message) gets a confidently wrong causal explanation. Suggest rewording to past-tense/hypothesis framing: "the pre-#64 claim was…, this test now guards the fix."

test/falcon_fiber_blocking_test.rb
```diff
@@ -0,0 +1,147 @@
...
+  # get a baseline of "normal" gaps as well as the gap(s) that occur while
+  # Duckling.parse is running.
+  TICKS_BEFORE_PARSE = 20
+  TICKS_AFTER_PARSE = 20
+
+  # Ticker gaps are allowed to exceed TICK_INTERVAL by this much before we
+  # consider them "blocked" — real async scheduling always has some jitter
+  # (OS scheduling, GC, etc.), so a tiny tolerance avoids false positives on
+  # a merely-slow tick.
+  NON_BLOCKING_TOLERANCE = 0.010 # 10ms
```
cpb: **[flake risk]** An absolute 10ms tolerance over 40 one-millisecond ticks is tight for shared CI runners — a single GC pause or OS scheduling hiccup >11ms in any one gap fails the test even though nothing blocked the reactor, and the failure message then asserts the (now-false) GVL-blocking explanation. Consider a proportional threshold instead — e.g. assert `max_gap` is well below the measured `parse_duration` (the actual blocking signature is a gap ≈ parse duration), which is robust to jitter.

AGENTS.md
```diff
@@ -25,7 +25,7 @@ not-yet-built piece as current, or vice versa, fix it in the same PR (see
...
-| `lib/duckling.rb` | Ruby module entrypoint (`Duckling` module, `Duckling::Error`). Will `require_relative "duckling/duckling"` to load the compiled native extension once it exists. |
+| `lib/duckling.rb` | Ruby module entrypoint. `require_relative "duckling/duckling"` loads the compiled native extension, which defines `Duckling::Native.parse` (the raw entrypoint — no thread spawn). `Duckling.parse` is then defined here as a thin Ruby-level wrapper, `Thread.new { Native.parse(...) }.value` (issue #64): thread-per-call dispatch so a calling Fiber on an `Async::Reactor` can yield to sibling Fibers while the GVL-released native call runs — see the "Rust/Magnus wiring" section's GVL-release bullet below for why a bare GVL release alone isn't sufficient. |
 | `lib/duckling/version.rb` | `Duckling::VERSION` constant — single source of truth for the gem version, read by `duckling.gemspec` and (once built) the release pipeline. |
 | `test/` | Minitest suite. `test_helper.rb` sets up the load path and requires `minitest/autorun`; test files currently follow `test_<name>.rb` / `class Test<Name> < Minitest::Test` naming. |
```
cpb: **[conventions]** This line's naming claim (`test_<name>.rb` / `class Test<Name>`) is the inverse of what every test file actually uses — all four, including this PR's new `falcon_fiber_blocking_test.rb` / `FalconFiberBlockingTest`, follow the `<name>_test.rb` / `<Name>Test` suffix style. Per this doc's own "Keeping this file current" rule, and since the PR already edits AGENTS.md two lines up, fix the description here in the same PR.

test/parse_error_stderr_test.rb
```diff
@@ -0,0 +1,39 @@
...
+# These tests assert the desired behavior: the exception still reaches the
+# caller exactly as before, with nothing written to stderr along the way.
+class ParseErrorStderrTest < Minitest::Test
+  def test_rescued_argument_error_writes_nothing_to_stderr
+    error = nil
+    _out, err = capture_subprocess_io do
+      Duckling.parse("tomorrow", locale: "zz-ZZ")
+    rescue ArgumentError => e
+      error = e
+    end
```
cpb, line 25: **issue:** this is a false pass, it needs to be in a fiber to hit the threaded path.

test/native_panic_test.rb
```diff
@@ -0,0 +1,72 @@
...
+  PANIC_PROBE = <<~'RUBY'
+    require "duckling"
+
+    Duckling.send(:remove_const, :Native)
+    Duckling.const_set(:Native, Duckling::PanickingNativeFake)
+
+    begin
+      Duckling.parse("tomorrow", locale: "en")
+      puts "NO_ERROR"
+    rescue => e
```
cpb, line 41: **issue:** lets add additional coverage for a panic probe, within a fiber scheduler

Issue #32, opened 2026-07-01
Explore serde_magnus (symbol keys) + Ruby pattern-matching Data factories as post-0.2.0 API


cpb:
## Goal

Explore replacing the manual-Magnus-hash-mapping API (shipped for 0.2.0) with symbol-keyed serde_magnus output consumed by Ruby `case/in` pattern-matching factory methods that construct `Data`-based value objects, instead of returning raw Hashes to callers.

## Background

The 0.2.0 implementation uses Option B from [`docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/serialization-options.md`](https://github.com/cpb/duckling/wiki/research-type-mapping-strategy-serialization-options) (manual `RHash`/`RArray` construction in Rust), chosen because plain `serde_magnus` produces an externally-tagged shape (`{"Time" => tv}`, `{"Single" => {...}}`, PascalCase `Grain` variants) that doesn't match the target Ruby Hash schema.

A PR #3 review comment proposed a different framing: instead of treating that externally-tagged shape as a defect to fix with serde container attributes, *keep* it — the tag keys (`"Time"`, `"Single"`, `"Naive"`, etc.) are exactly what Ruby's pattern matching wants to discriminate on. The only serde-side change needed is symbolizing keys (Ruby pattern matching / `deconstruct_keys` is built around Symbol-keyed Hashes). Ruby-side factory methods would then `case/in` pattern-match on that shape and construct `Data.define(...)`-based immutable value objects for `Entity`, `TimeValue::Single`, `TimeValue::Interval`, `TimePoint::Naive`, `TimePoint::Instant`, etc. — giving callers real value objects instead of Hashes to memorize and re-parse.

This has been added as "Option D" in `serialization-options.md` for reference. It was explicitly *not* adopted for 0.2.0 (which already shipped on Option B) — this issue tracks evaluating and potentially implementing it for a later release.

## Acceptance criteria

- [ ] Prototype: symbolize serde_magnus's default output keys for `Entity`/`DimensionValue`/`TimeValue`/`TimePoint`/`Grain` (minimal serde change — no `tag`/`rename_all` container attributes)
- [ ] Prototype: Ruby factory method(s) that `case/in` pattern-match on that shape and construct `Data`-based objects for each variant
- [ ] Compare ergonomics and code size against the current manual-mapping approach (Option B) for both the Rust bridge code and the Ruby-side API surface
[...]

PR #53, opened 2026-07-02
Cut over Duckling.parse to serde_magnus + Data value objects (issue #32)

cpb:
## Summary

Replaces 0.2.0's manual `RHash`/`RArray` construction (Option B) with `serde_magnus::serialize` + an in-place key-symbolizer + Ruby `case/in` factories building `Data` value objects (Option D), per issue #32. `Duckling.parse` now returns `Duckling::Entity`/`Duckling::TimeValue::{Single,Interval}`/`Duckling::TimePoint::{Naive,Instant}` `Data` objects instead of `Hash`es — a breaking change shipped as **0.3.0** (pre-1.0, so a minor bump rather than a major one).

- `ext/duckling/src/ruby_value.rs` (new): generic, GC-safe, in-place key symbolizer — works for every `DimensionValue` variant, not just Time.
- `ext/duckling/src/lib.rs`: native method moves to `Duckling::Native.parse` (raw symbol-keyed tagged Hash); `Duckling.parse` itself is now pure Ruby.
- `lib/duckling/entities.rb` (new): the `Data` classes and `case/in` factory.
- `bin/benchmark_parse` (new): `GC.stat`/`benchmark-ips` comparison tool, used to produce the numbers below.
- `docs/issue-32-serde-magnus-comparison.md` (new): full write-up — code size, shape differences found, a real GC-safety bug hit and fixed along the way, and the measured before/after numbers.

## Real numbers, not estimates

| Input | Objects/call (0.2.0 → 0.3.0) | i/s (0.2.0 → 0.3.0) |
|---|---|---|
| single time value | 38 → 136 (3.6x) | ~14.7k → ~8.2k (~1.8x slower) |
| time interval | 37 → 181 (4.9x) | ~565 → ~525 (~8% slower, within noise — dominated by duckling's own grammar cost either way) |
| email | 11 → 59 (5.4x) | noisy both sides, roughly comparable |

Full breakdown (including the native-layer vs. Ruby-factory-layer split) is in the comparison doc.

## Notable things found while building this

- A real GC-safety bug: the first version of the symbolizer staged deleted `Hash` entries in a Rust-native `Vec`, which segfaulted under load (values become invisible to MRI's conservative GC once off the Ruby object graph). Fixed by staging in a Ruby `RArray` instead; verified stable under `GC.stress` across 300 calls.
- `Entity` has no `dim` field in the wrapped Rust crate — 0.2.0's `:dim` was synthesized Rust-side; the new factory derives it from which pattern arm matched.
- `DimensionKind::Numeral` displays as `"number"`, not `"numeral"` — needed an explicit lookup table to preserve the existing `:number` vocabulary.
- `Grain::NoGrain` isn't reachable through any natural-language input we tried (used internally, normalized to `Second` before a resolved `Entity` is built) — pinned with a direct unit test of the lookup table instead of an end-to-end parse assertion.
- Bumped `ext/duckling`'s Rust edition 2021 → 2024 to resolve a formatting fight between this dev environment's commit hook (hardcoded `rustfmt --edition 2024`) and CI's plain `cargo fmt --check` (which reads the edition from `Cargo.toml`). Unrelated to the serde_magnus work itself, but needed for CI to pass.

## Test plan

- [x] `bundle exec rake` (standard + compile + test) green
- [x] `cargo fmt --manifest-path ext/duckling/Cargo.toml --check` clean
- [x] `cargo clippy --manifest-path ext/duckling/Cargo.toml -- -D warnings` clean
- [x] Manual `Duckling.parse`/`Duckling::Native.parse` sanity checks (single time value, interval, "now" → `:no_grain`, email dim)
- [x] `GC.stress = true` stress test (300 calls) confirming the symbolizer fix is solid
- [x] Before/after `bin/benchmark_parse` numbers captured in the comparison doc

Per the comparison doc's recommendation: ship it, given the ergonomics gain and that we're pre-1.0 — but the ~2-5x allocation increase is real and worth knowing about before merging.


cpb:
Holding this for benchmark-ips review before merging — please don't merge until the numbers in `docs/issue-32-serde-magnus-comparison.md` (and the follow-up camping-trip case) have been looked over.

cpb:
Superseded: its serde_magnus symbolizer + native-layer split has since been done independently and merged to main (#90, #93, soon #91 for Time). The Data-object factory layer this PR also built is being redesigned from scratch against that unified shape — tracked as a design proposal in #100, with this PR kept as prior art (see docs/issue-32-serde-magnus-comparison.md).

Issue #83, opened 2026-07-06
Zone-aware Instant-grain time arithmetic across DST transitions (blocked on upstream wafer-inc/duckling Context support)


cpb:
## Goal

Make `TimePoint::Instant`-grain time arithmetic (e.g. "in 4 months") DST-correct when the computed result lands on the other side of a DST transition from `reference_time:`.

## Current state

The wrapped `wafer-inc/duckling` crate's `Context` (its `src/resolve.rs`) is `chrono::FixedOffset`-only — it has no IANA timezone concept at all. Duration/relative arithmetic for `Instant` results is performed *inside the crate* against that single fixed offset, and the finished `DateTime<FixedOffset>` is what reaches this gem's `entity_to_ruby`. By the time the value crosses the Magnus boundary, there is nothing left for the gem to correct: re-offsetting the instant gem-side would change its wall-clock reading, not repair the arithmetic.

This is distinct from `TimePoint::Naive` results, which arrive as zone-less wall-clock values and *can* be resolved gem-side — that is what #75 (`reference_zone:`) covers. #75 also adds failing, skipped tests characterizing this Instant-grain limitation, so the gap stays visible in the suite.

## What fixing it would take

Upstream zone-aware `Context` support in `wafer-inc/duckling` (e.g. generic over `chrono::TimeZone` or built on `chrono-tz`), then threading a real zone through this gem's `build_context` (`ext/duckling/src/lib.rs`). Gem-side-only workarounds were considered and rejected in #75's scoping.

## Notes

Backlog: exploratory/upstream-dependent work — not scoped for automated pickup by babysitting loops. Related upstream-contribution track: #47.

PR #85, opened 2026-07-06
Add reference_zone: for DST-aware Naive-grain time resolution

### LINKED ISSUE #75: Make Naive-grain time values DST-aware via a reference_zone: parameter
## Goal

Resolve `TimePoint::Naive` values against a real IANA timezone (via a new `reference_zone:` parameter) so the offset applied is correct for each result's own date, instead of the single fixed offset applied uniformly today — which is wrong for any result whose date falls on the other side of a DST transition from `reference_time:`.

## Background

As shipped today (v0.3.0-rc1 on `main`), `Duckling::Native.parse` derives a single `chrono::FixedOffset` from `reference_time:`'s `utc_offset` (`build_context` in `ext/duckling/src/lib.rs`) and `resolve_naive` applies that one offset uniformly to every `TimePoint::Naive` result before it is returned as a Ruby `Time`. This is correct only when the reference time and every resolved result fall in the same DST regime. `wafer-inc/duckling` (the wrapped Rust crate, confirmed via its own source — `src/resolve.rs`'s `Context`) has no IANA timezone concept at all — it's `chrono::FixedOffset`-only, all the way down — so this gap can't be closed inside the wrapped crate; it needs a gem-side fix using a real timezone database.

Example of the bug this issue targets: `reference_time:` on 2013-01-15 in `America/New_York` (EST, `-05:00`); a Naive result resolves to "June 15" (should be EDT, `-04:00`, once DST is in effect) but today gets `-05:00` applied instead, since the gem only knows the reference's own numeric offset, not a real zone's transition rules.

## Acceptance criteria

- [ ] `Duckling.parse` accepts a new `reference_zone:` keyword (IANA zone name String, e.g. `"America/New_York"`), resolved via the `tzinfo` gem (new runtime dependency)
- [ ] When `reference_zone:` is given, each `TimePoint::Naive` result's offset is resolved against that zone for **that result's own date** (not the reference date's offset applied blindly) — a hill test asserts a Naive result landing after a real DST transition gets the post-transition offset
- [ ] When `reference_zone:` is omitted, behavior is unchanged from today (single fixed offset from `reference_time:` applied to every Naive result) — existing tests continue to pass unmodified
- [ ] `reference_zone:` given but `reference_time:` omitted — behavior is documented and tested. Choose one (delete the others):
  - [ ] Raise `ArgumentError`: a zone without an explicit anchor is likely a caller mistake; require both
  - [x] Anchor at the current time *in that zone* (equivalent of `reference_time: Time.now` reinterpreted through the zone), so "tomorrow" etc. resolve in local terms
  - [ ] Keep the default anchor (now, UTC — `Context::default()`) and use the zone only when applying offsets to Naive results
- [ ] `reference_time:`'s `utc_offset` disagrees with the zone's offset at that instant (e.g. a UTC `Time` with `reference_zone: "America/New_York"`) — behavior is documented and tested. Choose one (delete the others):
  - [x] Raise `ArgumentError` unless `reference_time.utc_offset == zone.utc_offset(reference_time)`: strict, catches confused callers early
  - [ ] Zone wins: keep `reference_time:`'s absolute instant but reinterpret its wall-clock/offset through the zone before building the parse `Context` (so relative expressions resolve in the zone's local frame regardless of how the `Time` was constructed)
  - [ ] No validation: `reference_time:`'s own offset anchors the parse `Context` exactly as today; `reference_zone:` is consulted only when resolving Naive results
- [ ] Create failing, skipped tests to support the documented known limitation about `TimePoint::Instant`-grain arithmetic remaining DST-imprecise across a transition boundary (#83)

## Out of scope

- `TimePoint::Instant`-grain arithmetic (e.g. "in 4 months") remains DST-imprecise across a transition boundary — the wrapped crate performs that arithmetic internally against a single `FixedOffset` before the result ever reaches this gem, and fixing it requires upstream zone-aware `Context` support, not a gem-side change. Tracked separately in #83; this issue only documents it via skipped tests.
- Auto-detecting a zone from an `ActiveSupport::TimeWithZone` passed as `reference_time:` — `reference_zone:` is explicit-only for now. Tracked separately in #84.

## Notes

The fixed-offset baseline this builds on is fully landed: `reference_time:` accepts a Ruby `Time` (with `#to_time` coercion for `TimeWithZone`/`DateTime` in `Duckling.parse`, `lib/duckling.rb`), and Naive results already come back as real Ruby `Time` objects carrying the reference's offset. This issue adds the per-date, zone-aware refinement on top of that shipped behavior.

cpb:
Closes #75.

## Summary

`Duckling.parse` now accepts `reference_zone:`, an IANA zone name (e.g.
`"America/New_York"`, resolved via the new `tzinfo`/`tzinfo-data` runtime
dependencies) that makes `TimePoint::Naive` (wall-clock) results DST-aware:
each such result's offset is resolved against the real zone for *that
result's own date*, instead of the single fixed offset from
`reference_time:` applied uniformly today.

`Duckling.parse`'s signature is now explicit named parameters (previously
`*args, **kwargs`).

## Behavior

- `reference_zone:` given, `reference_time:` given → each `:naive` result's
  offset is recomputed against the zone for its own resolved date (the
  DST-transition case this issue targets).
- `reference_zone:` omitted → unchanged from today: `reference_time:`'s
  fixed offset applies uniformly, exactly as before.
- `reference_zone:` given, `reference_time:` omitted → anchors at the
  current time *in that zone*.
- `reference_time:`'s `utc_offset` disagreeing with `reference_zone:`'s real
  offset at that instant → raises `ArgumentError`.
- A `:naive` result whose wall-clock time falls in a DST spring-forward gap
  or fall-back overlap for the given zone → raises `ArgumentError`, rather
  than leaking the underlying `TZInfo::PeriodNotFound`/`TZInfo::AmbiguousTime`.
- `TimePoint::Instant`-grain arithmetic (e.g. "in 5 months") is explicitly
  out of scope and remains DST-imprecise — that arithmetic happens inside
  the wrapped `wafer-inc-duckling` crate against a single `FixedOffset`
  before this gem ever sees the result. Documented via a skipped test
  (tracked separately in #83).

## How

All zone-awareness lives in Ruby (`lib/duckling.rb`) via `tzinfo` — neither
the wrapped crate nor this gem's own Magnus binding has any IANA/DST concept.
`ext/duckling/src/lib.rs` now tags every time-point Hash with a `:naive`
boolean (`true` for `TimePoint::Naive`, `false` for `TimePoint::Instant`),
since the two are otherwise indistinguishable once converted to a plain Ruby
`Time` — that's how `Duckling.parse` knows which results it may safely
reinterpret. `apply_reference_zone` gates on the entity's authoritative
`:dim == :time` tag rather than inferring it from `value[:type]` shape, and
raises loudly on any `value[:type]` it doesn't recognize instead of
silently leaving it untouched, so drift against the Rust side's shape can't
pass unnoticed.

The gem bundles `tzinfo-data` (not just `tzinfo`) so `reference_zone:`
resolution doesn't depend on the host having `/usr/share/zoneinfo` — this
keeps behavior identical on minimal/scratch container images, and lets a
stale IANA rule be picked up with a plain gem bump rather than a native
recompile.

## Alternative considered — resolved

Two entirely-different-seam alternatives were explored in parallel:

- **#88** (Rust-seam + Ruby `tzinfo` callback): pushed the per-date
  reinterpretation into Rust's `resolve_naive`, calling back into a Ruby
  `TZInfo::Timezone` via Magnus `funcall` — no `:naive` tag ever leaves
  Rust, but couples the compiled, precompiled-binary-shipping extension to
  `tzinfo`'s exact Ruby method signature.
- **#89** (fully-Rust via `chrono-tz`): compiled-in IANA tzdata, no Ruby
  `tzinfo` dependency at all, at the cost of that tzdata only being
  updatable via a `chrono-tz` release bump + new cross-compiled binaries,
  rather than a plain gem bump.

Both #88 and #89 are now closed in favor of this PR's approach. The
`tzinfo`/`tzinfo-data` split this PR ships with is deliberate, not
incidental: per `duckling.gemspec`, `tzinfo` (`~> 2.0`) resolves IANA
offsets, and `tzinfo-data` (`~> 1.2024`) is bundled alongside it as a
runtime dependency because `tzinfo` alone falls back to reading the host's
`/usr/share/zoneinfo`, which is absent from slim/scratch container images
and unpatchable at runtime when present. Bundling the tzdata keeps
`reference_zone:` working identically everywhere, and lets a consumer pick
up a tz-database revision with a plain `tzinfo-data` gem bump rather than a
native recompile.

## Test plan

- [x] Full suite green: `120 runs, 522 assertions, 0 failures, 0 errors, 6 skips`
- [x] Manually exercised the DST-transition, DST-gap/ambiguous-raises,
      Instant-grain (unaffected), zone-mismatch-raises, and unrecognized-shape
      cases directly against `Duckling.parse`


cpb:
## Hill ready for review

**To proceed:**
2. Open `/continue-pr 85` — that session will inspect CI and ask whether the hill is correctly specified.

cpb:
Two exploratory alternative architectures were built for comparison (both draft, not meant to merge):

- #88 — resolves `reference_zone:` at the Rust seam via a Ruby `TZInfo` callback (no `:naive` tag, but couples compiled Rust to `tzinfo`'s Ruby API)
- #89 — resolves `reference_zone:` entirely in Rust via `chrono-tz` (no `tzinfo` runtime dependency at all; see that PR for a few findings — return-type/error-type consistency gaps in this PR worth considering)

cpb:
Superseded by #102, which rebases and re-implements this work against #91's unified Time :value shape. Closing per #96's acceptance criteria.

PR #88, opened 2026-07-06
alt(75): resolve reference_zone: at the Rust seam via a Ruby TZInfo callback

cpb:
## Alternative architecture — comparison only, do not merge

One of two exploratory alternatives to **#85** (the shipped implementation of #75), built to compare against a third: **cpb/duckling#<chrono-tz-pr>** (chrono-tz, fully in Rust).

## What's different from #85

#85 tags every time-point Hash with a `:naive` boolean (Rust exposes which `TimePoint` variant produced it) and does the actual per-date zone reinterpretation in Ruby, via a post-processing loop over `Duckling.parse`'s return value.

This branch instead pushes the reinterpretation into Rust's `resolve_naive` — the function already only called from `TimePoint::Naive` match arms, so Rust already has the Naive/Instant distinction in hand right there. No `:naive` tag ever needs to leave Rust:

- `Native.parse` gains a `reference_zone:` kwarg (a plain IANA name string).
- After the off-GVL native call returns (GVL held again — safe to call Ruby here, per the existing GC-safety rules governing `parse_without_gvl`), `resolve_reference_zone` looks up `TZInfo::Timezone.get(name)` **once** via Magnus `funcall`, producing a Ruby `TZInfo::Timezone` object.
- `resolve_naive` calls back into that same object's `#local_time` per `TimePoint::Naive` result, getting back a `TZInfo::TimeWithOffset` (a `Time` subclass) with the correct offset for that result's own date.
- `lib/duckling.rb` shrinks to validation/anchoring only — no post-processing loop.

## Test plan

- [x] Full suite green: `116 runs, 518 assertions, 0 failures, 0 errors, 6 skips` (release profile — a fresh worktree's `bin/setup` seeds a dev-profile `.env.local` by default, which made an early run look 6x slower; rebuilding release-profile matched the baseline)
- [x] Manually verified DST-transition, Instant-grain (correctly unaffected), reference_time:/zone mismatch-raises, and interval `:from`/`:to` cases directly

## Known tradeoffs vs #85

- No new public API surface (no `:naive` key ever exposed) — a genuine win.
- But couples compiled Rust to `TZInfo::Timezone#local_time`'s exact Ruby method signature. This gem ships precompiled binaries (see AGENTS.md) — a `tzinfo` API change becomes a recompile-and-cross-compile-and-release problem here, versus a trivial patch release for #85's pure-Ruby version.


cpb:
Closing — superseded by #89 (chrono-tz), which gets the same seam-based benefit (no `:naive` tag needed) without this branch's downsides: no coupling of compiled Rust to `tzinfo`'s Ruby method signature, no runtime dependency on the system's zoneinfo data, and no per-call Ruby-callback overhead.

PR #89, opened 2026-07-06
alt(75): resolve reference_zone: entirely in Rust via chrono-tz

cpb:
## Alternative architecture — comparison only, do not merge

Second of two exploratory alternatives to **#85** (the shipped implementation of #75), for comparison against **#88** (Rust-seam + Ruby `tzinfo` callback).

## What's different from #85 and #88

Both #85 and #88 use Ruby's `tzinfo` gem as the source of truth for IANA offsets — #85 via a post-processing loop in Ruby, #88 via a Rust→Ruby callback per result. This branch instead adds `chrono-tz` (same GitHub org as `chrono`; compiled-in IANA tzdata, implements the same `chrono::TimeZone` trait `FixedOffset` already does) as a plain Cargo dependency, so **all** zone-awareness — validation, "anchor at current time in zone", and per-date `Naive` resolution — happens in Rust, with no Ruby callback and no `tzinfo` runtime dependency at all.

- `resolve_naive` keeps its *original* signature/return type (`Result<DateTime<FixedOffset>, Error>`) — when a zone is given, it resolves via `chrono-tz` then `.fixed_offset()` converts back to the same shape the non-zone branch already returns, so its callers need zero changes.
- `build_context` handles `reference_time:`/`reference_zone:` mismatch validation and "anchor at current time" directly, using `chrono-tz`'s `offset_from_utc_datetime`/`with_timezone`.
- `lib/duckling.rb` just forwards `reference_zone:` as a plain `String` — no `require "tzinfo"` anywhere.

## Test plan

- [x] Full suite green: `118 runs, 524 assertions, 0 failures, 0 errors, 6 skips` — matching the release-profile baseline exactly (no cross-language callback per result, unlike #88)
- [x] Manually verified DST-transition, Instant-grain (correctly unaffected), reference_time:/zone mismatch-raises, interval `:from`/`:to`, and an invalid zone name
- [x] DST-gap (spring-forward) and fall-back-ambiguous naive times now have regression tests pinning the `ArgumentError` contract (previously verified manually only)

## Findings favoring this version over #85

- **Portability**: `tzinfo` (as installed on #85) resolves via `TZInfo::DataSources::ZoneinfoDataSource` by default — i.e. it reads the **system's** `/usr/share/zoneinfo` at runtime; a minimal/scratch container without that directory needs the separate `tzinfo-data` gem. `chrono-tz`'s tzdata is compiled directly into the binary — no runtime file dependency, which fits this project's precompiled-binary-gem model (AGENTS.md) better.
- **No `:naive` tag as public API surface**: this implementation needs no equivalent of #85's `:naive` boolean on every time-point Hash, since the Naive/Instant distinction and the zone-aware resolution both happen in the same `resolve_naive` seam — the distinction never has to cross the FFI boundary.
- ~~Return type / error type consistency~~ — no longer a differentiator. #85's `bc673cf` ("Fix return-type and error-type inconsistencies in reference_zone: results") already fixed both issues this section used to cite (`TZInfo::TimeWithOffset` leaking as a `Time` subclass, and `TZInfo::InvalidTimezoneIdentifier` leaking instead of `ArgumentError`). Both implementations now return plain `Time` and raise `ArgumentError` consistently for bad zone names.

## Tradeoff cutting the other way, in #85's favor

- **`chrono-tz`'s IANA tzdata is compiled in, not runtime-updatable.** Its `tz` submodule is pinned to one fixed commit per release — confirmed at release 2025b (2025-03-22) as of `chrono-tz` 0.10.4, the version this branch's `Cargo.lock` uses, with no subsequent bump seen in that upstream repo's later history. A real-world DST-rule change (precedented by Yukon's 2020 switch to permanent UTC-7, encoded in IANA's `northamerica` file as a `Zone` continuation line dropping its `Rule` reference) only reaches this gem after `chrono-tz` cuts a new release, this crate's `Cargo.toml` constraint is bumped, and a new native-extension release ships with new cross-compiled binaries. #85's `tzinfo` can pick up new IANA data via a plain `tzinfo-data` gem bump instead, no recompile required. (Not asserting any specific jurisdiction — e.g. British Columbia — has made such a change; stating the mechanism and lag, which is real regardless.) See `AGENTS.md`'s Rust/Magnus wiring section and the comment in `ext/duckling/Cargo.toml` for the full note.

## Known limitation (shared by all three alternatives)

None of these can fix **#83** (`TimePoint::Instant`-grain arithmetic, e.g. "in 5 months", remaining DST-imprecise). Confirmed by reading the wrapped `wafer-inc-duckling` crate's own source: its `Context.reference_time` is hardcoded to `DateTime<FixedOffset>`, and its arithmetic (`dimensions/time/mod.rs`) is ~15+ call sites of `checked_add_signed`/`checked_sub_signed` directly on `FixedOffset` values with no zone identity to re-derive an offset from. Fixing #83 requires patching (or forking) that crate itself, not just adding a zone-aware dependency to our own binding.


cpb:
closing in favour of:

because of the more timely maintenance of the tzinfo-data gem compared to chrono-tz

Issue #91, opened 2026-07-06
Migrate DimensionValue::Time onto the unified serde_magnus tagged shape (test-first)


cpb:
## Goal

Migrate `DimensionValue::Time` onto the same unified, externally-tagged `serde_magnus` shape adopted for the other 13 dimensions in #90 — a breaking change to `Time`'s already-shipped `:value` shape, done test-first given the regression risk.

## Background

#90 keeps `DimensionValue::Time` on its existing bespoke, flattened conversion (`time_value_to_ruby`/`time_point_to_ruby`/`resolve_naive` in `ext/duckling/src/lib.rs`) specifically because Time is the one dimension already shipped and depended on. This issue closes that gap for full consistency: every dimension should return its `:value` via the same tagging convention.

**Confirmed technical constraint** (verified directly against the `serde_magnus-0.11.0` source, `src/ser/serializer.rs`): its `Serializer` impl has no escape hatch — `serialize_newtype_struct`/`serialize_newtype_variant`/every other method bottoms out in a primitive `serialize_*` call (`serialize_str`, `serialize_i64`, etc.) with no way to inject a raw Magnus `Value` mid-serialization. The wrapped crate's `TimePoint`/`TimeValue` (`duckling-0.4.0/src/types.rs`) derive plain `serde::Serialize`, and their `DateTime<FixedOffset>`/`NaiveDateTime` fields serialize via chrono's own `serde` support — an RFC3339-ish **String**, not a Magnus `Time`. Generically serializing `entity.value` for a Time entity would therefore silently turn `:value`'s datetime into a Ruby `String`, breaking the documented, load-bearing guarantee that "`:value` is always a real Ruby `Time`, never a string" (see `Duckling::Native.parse`'s docstring in `lib.rs` and AGENTS.md's "Public and internal APIs" section).

Separately, `TimePoint::Naive`'s bare `NaiveDateTime` has no offset at all — resolving it into an absolute time (`resolve_naive`) is real business logic requiring the parse-time reference-offset context, not something any serialization framework can derive structurally.

## Design

Reuse the "generic-serialize-then-patch-known-leaves" pattern #90 establishes for `Grain` casing (`TimeGrain`/`Duration`), extended to cover Time's two problem leaf types:

[...]

Issue #46, opened 2026-07-02
Implement remaining DimensionValue variants beyond Time


cpb:
## Goal

Add Ruby mapping support for the 13 `DimensionValue` variants beyond `Time` (Numeral, Ordinal, Temperature, Distance, Volume, Quantity, AmountOfMoney, Email, PhoneNumber, Url, CreditCardNumber, TimeGrain, Duration), so `dims:` can request dimensions other than `["time"]` without raising `ArgumentError`.

## Background

0.2.0 shipped `entity_to_ruby` handling only `DimensionValue::Time` — the default `dims: ["time"]` means callers never hit this gap, but passing any other dimension string explicitly currently raises `ArgumentError` from the dim parser. This was tracked as an open item in [`docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/plans/00-pr2-roadmap.md`](https://github.com/cpb/duckling/wiki/plans-stale-00-pr2-roadmap) ("Post-PR-#2 follow-on work" #3) and `02-ruby-api-design.md` ("Open Questions" — "Non-Time dimensions in 0.2.0").

[`docs/1-ship-duckling-gem-time-extraction-via-magnus-wafer/research/type-mapping-strategy/serialization-options.md`](https://github.com/cpb/duckling/wiki/research-type-mapping-strategy-serialization-options) already documents the serde shape analysis showing manual mapping (Option B, the approach 0.2.0 shipped with) is needed for all variants, not just Time — this issue is executing that groundwork for the remaining 13.

## Acceptance criteria

- [ ] Each of the 13 remaining `DimensionValue` variants has a corresponding manual Magnus mapping arm in the Rust bridge (mirroring the pattern already used for `Time`)
- [ ] `Duckling.parse(text, dims: [...])` accepts any of the 14 valid dimension name strings without raising, matching pyduckling's dimension-string behavior already documented in `pyduckling-reference.md`
- [ ] Test coverage for at least one representative case per new dimension (e.g. `"$42.50"` for AmountOfMoney, `"5 pounds of sugar"` for Quantity)
- [ ] Existing `Time`-only behavior and hill tests are unaffected

## Out of scope

- The serde_magnus/Data-object API exploration (#32) — this issue extends the currently-shipped manual-mapping approach; if #32 lands first, coordinate rather than duplicate work

## Notes

[...]

PR #93, opened 2026-07-06
Populate :value for all 13 non-Time dimensions via serde_magnus's generic symbolizer

### LINKED ISSUE #90: Adopt serde_magnus's generic symbolizer (no Data factory) to close all 13 non-Time DimensionValue gaps
## Goal

Close all 13 remaining `DimensionValue` gaps from #46 in one PR by adopting a scaled-down version of PR #53's generic `serde_magnus`-based conversion — **without** PR #53's Ruby `Data`-object factory layer.

`DimensionValue::Time`'s migration onto this same mechanism is tracked separately in #91: it touches already-shipped, tested behavior and has a real technical wrinkle (see #91), so it gets its own test-first hill rather than riding along here. This issue's generic mechanism doesn't touch Time at all — `entity_to_ruby`'s existing `Time` arm stays exactly as-is until #91 lands.

## Background

A spike (see #46's discussion) measured the real cost of `serde_magnus`'s generic symbolizer (`ruby_value.rs::symbolize_keys_in_place` on PR #53's branch) in isolation from the Data-factory layer, using the current `benchmark/parse_benchmark.rb` harness:

| input | manual mapping | serde_magnus (no Data factory) | absolute added latency |
|---|---|---|---|
| single time value | 68.7 μs/call | 93.6 μs/call | +24.9 μs |
| time interval | 1,672 μs/call | 1,727 μs/call | +54.9 μs |
| email dim | 17.9 μs/call | 27.6 μs/call | +9.7 μs |

Negligible for this gem's actual use cases (e.g. logging ex-ante/ex-post signals around an LLM generation, where the generation call itself dominates by 4-6 orders of magnitude) and for anything but the cheapest, simplest inputs — worth adopting given it eliminates 13 discrete hand-written match arms in favor of one generic mechanism. Supersedes the manual-mapping plan from #86 (closed).

## Design

- **`DimensionValue::Time` is untouched here.** `entity_to_ruby`'s existing `Time` arm (flattened hash via `time_value_to_ruby`/`time_point_to_ruby`/`resolve_naive`) stays exactly as-is in this issue — zero regression risk for the one dimension already shipped and tested. The generic path below only applies to the other 13 variants (today's `_ => {}` catch-all). See #91 for Time's own migration onto this mechanism.
- **Port `symbolize_keys_in_place`** from PR #53's `ext/duckling/src/ruby_value.rs` (already GC-safety-audited: stages deleted `Hash` keys in a Ruby `RArray`, never a Rust-native `Vec<Value>` — see the GC-safety incident documented there and in this repo's own past segfault history). Recursive, in-place, works through `Hash` and `Array` at any depth.
- **Add Cargo deps**: `serde = "1"`, `serde_magnus = "0.11"` (current crates.io version, verify at implementation time).
- **For each non-Time variant**: `serde_magnus::serialize(ruby, &entity.value)?` produces an externally-tagged, single-key Hash (e.g. `{"Numeral" => 42.0}`, `{"Url" => {"value" => "...", "domain" => "..."}}`). Symbolize keys via the ported helper, then **unwrap the single outer tag** (redundant with the entity's existing `:dim` key, which is computed independently via `entity.value.dim_kind().to_string()` — unaffected by any of this, keep as-is) so `:value` is the tag's inner payload directly, not a redundant wrapper.
- **Known gotcha to handle explicitly** (found while researching this): the wrapped crate's `Grain` enum (`#[derive(... serde::Serialize)]` in `duckling-0.4.0`, no `rename_all`) serializes variant names verbatim in PascalCase (`"Second"`, `"NoGrain"`) — inconsistent with the existing `:second`/`:no_grain` symbol convention already used for `Time`'s grain field via the crate's own `Grain::as_str()` (see current `time_point_to_ruby`). This affects exactly two variants here: `TimeGrain(Grain)` and `Duration { grain: Grain, .. }`. For these two, don't trust serde's raw output for the grain field/value — substitute `grain.as_str()` (already-available from the match arm's typed binding) so grain symbols stay consistent across every dimension that has one. Everything else generic-converts cleanly with no casing exceptions.
  - **This "generic-serialize-then-patch-known-leaves" pattern is the one #91 reuses** for Time's own Grain fields and (more importantly) its `DateTime`/`NaiveDateTime` fields — establish it cleanly here since it's load-bearing for that follow-up.
- **Stress-test GC safety**: verify with `GC.stress = true` across ~300 `Duckling.parse` calls spanning all 13 dims, mirroring PR #53's own verification of this exact symbolizer (it found and fixed a real segfault there — see that PR's description).

## Acceptance criteria

- [ ] Each of the 13 remaining `DimensionValue` variants (Numeral, Ordinal, Temperature, Distance, Volume, Quantity, AmountOfMoney, Email, PhoneNumber, Url, CreditCardNumber, TimeGrain, Duration) populates a sensible `:value` for its entity, via the generic mechanism above
- [ ] `TimeGrain`/`Duration`'s grain field/value renders as the existing lowercase-snake_case symbol convention (`:second`, `:no_grain`, etc.), not serde's raw PascalCase
- [ ] `DimensionValue::Time`'s existing behavior and all existing hill tests are unaffected — no diff to `time_value_to_ruby`/`time_point_to_ruby`/`resolve_naive` or their call sites (that migration is #91)
- [ ] Test coverage for at least one representative case per new dimension (e.g. `"$42.50"` for AmountOfMoney, `"5 pounds of sugar"` for Quantity, per #46's original criteria)
- [ ] `GC.stress = true` stress test across all 13 dims plus Time, confirming the ported symbolizer is solid under this codebase's own GVL-release/panic-catch machinery (this repo's `entity_to_ruby` call path differs from PR #53's — that PR predates the GVL-release/`catch_unwind`/`reference_time` work on current `main`, so don't assume its call-site wiring transfers unmodified)
- [ ] `cargo clippy -- -D warnings` / `cargo fmt --check` clean on `ext/duckling/`
- [ ] AGENTS.md's "Public and internal APIs" section updated to describe the new `:value` shape for non-Time dimensions

## Out of scope

- PR #53's Ruby `Data`-object factory layer (`lib/duckling/entities.rb`) and the `Duckling.parse` → `Data` object breaking change — that's issue #32's separate, still-undecided scope. This issue only adopts the native-layer symbolizer, keeping `Duckling.parse`'s return shape a `Hash` exactly as today (additive change: non-Time entities gain a populated `:value` key they previously lacked; nothing existing changes shape)
- Re-deriving `:dim` from the serde tag (PR #53 needed this since its `Entity` Data-factory has no native `dim` field to fall back on) — current `main` already computes `:dim` independently via `dim_kind()`, unaffected by any of this
- Migrating `DimensionValue::Time` itself onto this mechanism — tracked in #91, gated on this issue landing first

## References

- #46 (parent issue — closes when this + #91 both land)
- #86 (closed — superseded manual-mapping plan)
- #91 (follow-up: migrates Time onto this same mechanism, test-first)
- PR #53 / issue #32 (`docs/issue-32-serde-magnus-comparison.md` on that branch — origin of `symbolize_keys_in_place` and the GC-safety fix)

cpb:
Closes #90.

Adopts a scaled-down version of PR #53's generic `serde_magnus` conversion — without its Ruby `Data`-object factory — to close all 13 remaining `DimensionValue` gaps from #46 in one PR. Additive change only: non-Time entities gain a populated `:value` key; nothing existing changes shape. `DimensionValue::Time` is byte-for-byte untouched (its migration is #91).

## The `:value` shape contract

Every non-Time dimension's `:value` is serde's externally-tagged representation with all Hash keys symbolized — the PascalCase tag key is kept uniformly at every enum layer, so consumers see one consistent tagged shape rather than a mix of unwrapped and tagged layers (per review, replacing an earlier design that unwrapped the outer key everywhere except `MeasurementValue`):

| Dimension | `:value` shape |
|---|---|
| number | `{Numeral: 42.0}` |
| ordinal | `{Ordinal: 3}` |
| email | `{Email: "duck@example.com"}` |
| phone-number | `{PhoneNumber: "6501234567"}` (normalized, separators stripped) |
| url | `{Url: {value:, domain:}}` |
| credit-card-number | `{CreditCardNumber: {value:, issuer:}}` (number normalized) |
| duration | `{Duration: {value:, grain: :hour, normalized_seconds:}}` |
| quantity | `{Quantity: {measurement:, product:}}` (`product` an explicit `nil` when absent) |
| time-grain | `{TimeGrain: :second}` |
| temperature / distance / volume / amount-of-money | `{Temperature: {Value: {value:, unit:}}}` etc.; intervals as `{AmountOfMoney: {Interval: {from:, to:}}}` |

Grain leaves are the one deliberate departure from raw serde output: they render as the shipped `Grain::as_str()` symbol convention (`:second`, `:no_grain`) rather than serde's PascalCase variant names, matching the Time dimension's existing `:grain` values.

## How it works

- `serde = "1"` + `serde_magnus = "0.11"` added to `ext/duckling/Cargo.toml` (serde_magnus 0.11.0 requires magnus `^0.8.1`, resolving cleanly with the existing `magnus = "0.8"` pin).
- `symbolize_keys_in_place` ported **verbatim** (GC-safety doc comments included) from PR #53's branch into a new `ext/duckling/src/ruby_value.rs` module, alongside a new `serialize_symbolized` helper: serde_magnus-serialize → symbolize all Hash keys in place, preserving the tagged structure verbatim.
- `entity_to_ruby`'s tail becomes a `match`: the `Time` arm is character-identical to before (zero diff to `time_value_to_ruby`/`time_point_to_ruby`/`resolve_naive`); `Duration` generic-serializes then patches the `:grain` leaf inside its tagged payload (the pattern #91 reuses for Time); `TimeGrain`, whose grain *is* the whole payload, builds its tagged one-pair Hash directly; all other variants go through `serialize_symbolized` unmodified.

## Acceptance criteria

- ✅ Each of the 13 variants populates a sensible `:value` (pinned per-dimension in `test/duckling_parse_dimensions_test.rb`, shapes verified empirically against the live parser before hardcoding)
- ✅ Serde's PascalCase tag key kept uniformly across all 13 dimensions (review feedback on the earlier unwrap-the-outer-key design)
- ✅ TimeGrain/Duration grains render as `:second`/`:no_grain`-style symbols, not PascalCase
- ✅ Time untouched — no diff to its helpers or call sites; all existing tests pass unmodified
- ✅ ≥1 representative test per dimension, plus shape-pinning extras for the measurement `Interval` form and Quantity's explicit `product: nil`
- ✅ `GC.stress = true` stress test across all 13 dims + Time single/interval (`test/duckling_gc_stress_test.rb`). The full ~300-call adoption-time verification (`DUCKLING_GC_STRESS_ITERATIONS=20`) ran green (~2 min); the checked-in default is 2 passes (~12 s) since `GC.stress` forces a collection at every allocation, so a single pass already interposes a GC at every conversion allocation site — keeping the default `bundle exec rake` under this repo's own too-slow threshold that got the benchmark suite excluded
- ✅ `cargo clippy -- -D warnings` and `cargo fmt --check` clean
- ✅ AGENTS.md "Public and internal APIs" + "Test guide" updated

## Test results

`bundle exec rake` (StandardRB + compile + Minitest): 127 runs, 595 assertions, 0 failures, 0 errors, 5 skips (pre-existing). All four CI jobs green.

https://claude.ai/code/session_01R4J8paCGV3dJREUKczmu1z


AGENTS.md
```diff
@@ -57,6 +57,7 @@ emails, etc.) without running a separate HTTP service.
...
... other 13 dimensions get their `:value` generically: `serde_magnus` serializes the crate's `DimensionValue`, the redundant externally-tagged outer key (already carried by `:dim`) is unwrapped, and all Hash keys are symbolized in place (`ext/duckling/src/ruby_value.rs`). Resulting shapes: bare scalars for number (Float), ordinal (Integer), email/phone-number (String — normalized, separators stripped); symbol-keyed Hashes for url (`{value:, domain:}`), credit-card-number (`{value:, issuer:}` — number normalized), duration (`{value:, grain:, normalized_seconds:}`), quantity (`{measurement:, product:}` — `product` is an explicit `nil` when absent, serde emits `Option::None` as a present key); and grain symbols (`:second`, `:no_grain`, …) for time-grain and duration's `:grain`, patched from serde's raw PascalCase variant names to match Time's `Grain::as_str()` convention. The five measurement dimensions (temperature, distance, volume, quantity's `:measurement`, amount-of-money) keep serde's externally-tagged `MeasurementValue` representation as a deliberate choice: a PascalCase-keyed one-pair Hash, `{Value: {value:, unit:}}` or `{Interval: {from: {value:, unit:}, to: {value:, unit:}}}`.
```
cpb: **polish:** I would like to maintain the PascalCase key uniformly, rather than just in the MeasurementValue case.

ext/duckling/src/lib.rs
```diff
@@ -481,11 +486,40 @@ fn entity_to_ruby(ruby: &Ruby, entity: &Entity, offset: FixedOffset) -> Result<V
...
+            value.aset(ruby.to_symbol("TimeGrain"), ruby.to_symbol(grain.as_str()))?;
+            h.aset(ruby.to_symbol("value"), value)?;
+        }
+        // Generic-serialize-then-patch-known-leaves (the pattern issue #91
+        // reuses for Time): serde's output is right except the :grain leaf
+        // inside the tagged payload, which gets the same symbol treatment as
+        // TimeGrain above.
+        DimensionValue::Duration { grain, .. } => {
+            let value = serialize_symbolized(ruby, &entity.value)?;
+            if let Some(tagged) = RHash::from_value(value) {
```
cpb: Duration arm's nested if-lets silently skip the :grain symbol patch if the serialized shape ever drifts, and the serialize-then-patch is more fragile than building the 3-field hash directly — which is exactly how the sibling TimeGrain arm is now written.

Failure scenario: A future duckling crate upgrade that renames/nests Duration's grain field (or changes its serde representation) makes `tagged.aref(:Duration)` or the field patch miss silently: Duration entities ship grain as serde's raw PascalCase "Day" string while Time entities use :day, with no error. Duration has exactly three scalar fields (value: i64, grain: Grain, normalized_seconds: i64), so a direct build removes both silent-skip paths; the arm's comment defends the pattern as what issue #91 reuses for Time, so this is a deliberate tradeoff to re-weigh.

ext/duckling/src/lib.rs
```diff
@@ -481,11 +486,40 @@ fn entity_to_ruby(ruby: &Ruby, entity: &Entity, offset: FixedOffset) -> Result<V
...
+        // "NoGrain"); the shipped convention is `Grain::as_str()` symbols
+        // (:second, :no_grain) — see time_point_to_ruby. For TimeGrain the
+        // grain *is* the whole tagged payload, so the hash is built directly
+        // rather than serialized-then-patched.
+        DimensionValue::TimeGrain(grain) => {
+            let value = ruby.hash_new();
+            value.aset(ruby.to_symbol("TimeGrain"), ruby.to_symbol(grain.as_str()))?;
+            h.aset(ruby.to_symbol("value"), value)?;
+        }
+        // Generic-serialize-then-patch-known-leaves (the pattern issue #91
```
cpb: New comments narrate bare issue/PR numbers, violating AGENTS.md 'Code comment conventions'.

AGENTS.md: 'Prefer explaining the durable why … over narrating "issue #N added this" … link to its permanent home on the wiki rather than a bare issue number, which reads as noise once the issue is closed.' Violations: ext/duckling/src/lib.rs:506 ('the pattern issue #91 reuses for Time'), test/duckling_gc_stress_test.rb:10 and :37 (an empirical result — 'PR #53's … caught a real segfault' — is exactly the case the rule says to link to the wiki), test/duckling_parse_dimensions_test.rb:6 ('(issue #90)').

ext/duckling/Cargo.toml
```diff
@@ -10,4 +10,6 @@ crate-type = ["cdylib"]
 magnus = { version = "0.8", features = ["chrono"] }
 duckling = "0.4"
 chrono = "0.4"
-rb-sys = { version = "*", default-features = false, features = ["stable-api-compiled-fallback"] }
+serde = "1"
+serde_magnus = "0.11"
+rb-sys ={ version = "*", default-features = false, features = ["stable-api-compiled-fallback"] }
```
cpb: The rb-sys line was gratuitously reformatted to `rb-sys ={ version = ...` (space moved to the wrong side of `=`) while adding the serde deps.

Failure scenario: Valid TOML with nothing lost, but it's an unremarked edit to a line the change didn't need to touch — diff noise now, and churn again when any Cargo.toml formatter later fixes it. cargo fmt doesn't cover Cargo.toml, so nothing in CI catches it. Restore the original spacing.

ext/duckling/src/ruby_value.rs
```diff
@@ -0,0 +1,77 @@
...
+/// `RArray` variable for the duration of this function).
+pub fn symbolize_keys_in_place(ruby: &Ruby, value: Value) -> Result<(), Error> {
+    if let Some(hash) = RHash::from_value(value) {
+        let keys = ruby.ary_new();
+        hash.foreach(|k: Value, _v: Value| {
+            keys.push(k)?;
+            Ok(ForEach::Continue)
+        })?;
+
+        for i in 0..keys.len() as isize {
```
cpb, line 34: symbolize_keys_in_place pays a delete+reinsert round-trip for every key, but serde_magnus 0.11 already emits Symbol keys for struct fields — only enum-tag layers produce String keys, so the majority of entries are deleted and reinserted unchanged.

Failure scenario: Two hash-table mutations per key on the per-entity conversion path, mostly no-ops: for a non-String key the code deletes an entry and reinserts the identical key/value when all it needs is to recurse into the value. Branch on String::try_convert first and, for non-String keys, read via aref and recurse without touching the hash structure.

test/duckling_gc_stress_test.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+REFERENCE_TIME = Time.new(2013, 2, 12, 4, 30, 0, "-02:00") unless defined?(REFERENCE_TIME)
```
cpb: New tests extend existing duplication instead of consolidating: a 9th copy of the REFERENCE_TIME constant, and the gc-stress test open-codes parse+lookup while the sibling new file defines a private entity_for helper that ~8 older files also inline.

Failure scenario: The `unless defined?` guard couples correctness to file load order, and a future change to the fixture instant or entity-lookup idiom must now be applied in two new places plus eight old ones. test_helper.rb (already required everywhere) is the single correct home for both.

test/duckling_gc_stress_test.rb
```diff
@@ -0,0 +1,54 @@
...
+    ["from 3pm to 5pm", "time"]
+  ].freeze
+
+  # GC.stress forces a collection at every allocation, so a single pass over
+  # CASES already interposes a GC between every allocation the conversion
+  # makes — extra passes add heap-layout variety, not new coverage. Each pass
+  # costs ~6s (every allocation pays a full-heap mark), so the default stays
+  # small enough for every `bundle exec rake`; bump the env var to re-run the
+  # ~300-call adoption-time verification (20 passes, mirroring PR #53's).
+  ITERATIONS = Integer(ENV.fetch("DUCKLING_GC_STRESS_ITERATIONS", 2))
```
cpb: The default of 2 GC-stress passes spends ~6s of every `bundle exec rake` on a pass that, by the test's own comment, adds 'heap-layout variety, not new coverage'.

Failure scenario: The whole suite now runs ~17s, and this one test is ~12s of it. A default of 1 pass preserves the guarded failure mode (GC forced at every allocation already interposes a collection at every conversion site) and halves the added runtime; the 20-pass verification stays available via DUCKLING_GC_STRESS_ITERATIONS.

cpb (commented):
Automated review findings for the autofixer to address. One finding (README.md) isn't inline because README.md isn't touched by this diff:

**README.md:70** — README still documents the pre-PR contract: ':value (Hash) — present only for the :time dimension today', and lines 95-101 say the other 13 dimensions 'currently have no :value key' — both now false, and the PR updates only AGENTS.md.

Failure scenario: A user follows the shipped README and writes `entity.key?(:value) ? handle_time(entity) : skip` to discriminate time entities; after this PR every entity carries :value, so handle_time receives {Numeral: 42.0} or {Email: "..."} and raises or corrupts downstream. The '(Hash)' type annotation is also wrong for the new tagged scalar payloads.

cpb:
Pushed fixes for all 7 findings in b41be80:

- README.md updated to document the new populated `:value` contract for all 13 non-Time dimensions (the README-only finding, since it wasn't in the diff).
- `Duration`'s `:value` is now built directly (matching the `TimeGrain` arm) instead of serialize-then-patch.
- Restored the original `rb-sys` Cargo.toml spacing.
- `symbolize_keys_in_place` no longer deletes+reinserts already-Symbol keys.
- Consolidated `REFERENCE_TIME`/`entity_for` into `test_helper.rb`.
- `DUCKLING_GC_STRESS_ITERATIONS` default dropped from 2 to 1.
- Reworded comments that narrated bare issue/PR numbers.

`bundle exec rake` (standard + compile + test) and `cargo clippy -- -D warnings` / `cargo fmt --check` are all green.

---

test/duckling_test.rb
```diff
@@ -5,11 +5,6 @@
 
 VALID_GRAINS = %i[second minute hour day week month quarter year].freeze
 
-# Matches the reference time used throughout the pyduckling / wafer-inc-duckling
-# corpora (2013-02-12T04:30:00-02:00), so relative expressions resolve to fixed,
-# assertable values instead of drifting with the real clock.
-REFERENCE_TIME = Time.new(2013, 2, 12, 4, 30, 0, "-02:00")
```
cpb, line 11: **polish:** there is more reference time references we can dry up in this test suite I believe

PR #94, opened 2026-07-06
test(75): shared reference_zone: test suite for #85/#89

cpb:
Extracted base branch for issue #75's `reference_zone:` work.

## Why this PR exists

PR #85 and PR #89 are two competing implementations of `reference_zone:`
(Ruby/`tzinfo` post-processing vs. entirely-in-Rust `chrono-tz`) — see #75.
Both were built on the same hill-derived test suite and, since then, both
independently added the identical DST-gap/ambiguous-time regression test.
Splitting the shared test suite into its own branch/PR lets each
implementation PR be rebased on top of it, so reviewing #85 vs. #89 only
shows the actual implementation differences, not duplicated test content.

## What's in it

- The complete hill-derived `reference_zone:` test suite (`test/duckling_test.rb`):
  zone-omitted-unchanged, DST-transition-offset, instant-grain-dst-limitation
  (skipped, tracked separately in #83), reference-time-zone-mismatch,
  zone-without-reference-time-anchors-at-current-time.
- One new test pinning the DST-gap/ambiguous-time `ArgumentError` contract
  (a naive time landing in a spring-forward gap or fall-back overlap must
  raise `ArgumentError`, not an implementation-specific exception) — added
  independently to both #85 and #89 this session; consolidated here as the
  single shared source of truth.

## Note

These tests currently **fail** against `main` as merged here, by design —
`Duckling.parse` doesn't support `reference_zone:` at all yet on `main`.
That's expected: this PR only lands the test suite; #85 or #89 (whichever
is chosen) is what makes it pass. Not intended to be merged to `main` on
its own before that decision is made — it exists so both implementation
PRs can share one base.


cpb:
closing in favour of merging with:

PR #95, opened 2026-07-06
alt(75): resolve reference_zone: via ActiveSupport::TimeZone/TimeWithZone

cpb:
Alternative to #75, alongside #85 (plain `tzinfo`) and #89 (entirely-in-Rust
`chrono-tz`). Same base branch (#94's shared test suite), same behavioral
contract, different implementation: resolves `reference_zone:` via
`ActiveSupport::TimeZone` and returns `ActiveSupport::TimeWithZone` (not a
plain `Time`) for reinterpreted results.

## Summary

`Duckling.parse` accepts `reference_zone:`, an IANA zone name resolved via
`ActiveSupport::TimeZone` (new `activesupport` runtime dependency), making
`TimePoint::Naive` (wall-clock) results DST-aware: each such result's offset
is resolved against the real zone for *that result's own date*.

## Behavior

Identical contract to #85/#89 (pinned by the shared test suite, unmodified
in this PR):

- `reference_zone:` + `reference_time:` given → each `:naive` result's
  offset is recomputed against the zone for its own resolved date.
- `reference_zone:` omitted → unchanged from today.
- `reference_zone:` given, `reference_time:` omitted → anchors at the
  zone's current time.
- `reference_time:`'s `utc_offset` disagreeing with the zone's real offset
  at that instant → `ArgumentError`.
- A `:naive` result landing in a DST gap/ambiguous overlap → `ArgumentError`.
- `TimePoint::Instant`-grain arithmetic stays out of scope (issue #83).

## The actual pitch: `ActiveSupport::TimeWithZone` as the return type

This is the one real, callable-visible difference from #85/#89, not just an
implementation detail. A reinterpreted `:naive` result comes back as an
`ActiveSupport::TimeWithZone`, which carries the zone's own identity forward
(`#time_zone`, `#period`) rather than just a numeric UTC offset. It's still
a drop-in `Time` for callers that don't care: ActiveSupport patches
`TimeWithZone#is_a?`/`#kind_of?` and `Time.===` to report `true` for it, so
`kind_of?(Time)`, `==` against a plain `Time`, and delegated methods
(`#utc_offset`, `#to_date`, ...) all work — confirmed the shared test
suite's plain-`Time` assertions pass against it unmodified.

## Two gotchas found and fixed along the way

1. **`reference_time:`'s existing coercion guard breaks under
   ActiveSupport.** The pre-existing `!reference_time.is_a?(Time)` guard
   (meant to coerce `TimeWithZone`/`DateTime`/etc. via `#to_time` before
   `Native.parse`'s strict Magnus `kind_of?(Time)` check) is silently
   defeated by the same ActiveSupport patch mentioned above:
   `reference_time.is_a?(Time)` returns `true` for a `TimeWithZone` even
   though Magnus's C-level check doesn't see it that way, so the coercion
   never fires and a raw `TimeWithZone` reaches `Native.parse` and raises
   `TypeError`. Reproduced live (including against #85's own branch, which
   has the identical guard — this is a live bug there too, not specific to
   this PR). Fixed here by checking `instance_of?(Time)` instead, which
   isn't patched.
2. **`ActiveSupport::TimeZone#local`'s own DST resolution can't be used for
   `:naive` reinterpretation.** It silently *repairs* a spring-forward-gap
   time (advances an hour, retries) and silently resolves fall-back
   ambiguity to the daylight period — neither raises, confirmed by reading
   `TimeWithZone#get_period_and_ensure_valid_local_time` and
   `TimeZone#period_for_local`'s own source. To satisfy this gem's
   `ArgumentError` contract, `reinterpret_time_point!` calls
   `TZInfo::Timezone#period_for_local` directly (via `TimeZone#tzinfo`) with
   `dst: nil`, then hands the resolved period straight to `TimeWithZone.new`
   so ActiveSupport's auto-repair path is never reached.

## Dependency weight

`activesupport` is a much heavier runtime dependency than `tzinfo` alone.
Requires the full top-level `require "active_support"`, not just the
narrower `require "active_support/time"` — the latter leaves
`ActiveSupport.utc_to_local_returns_utc_offset_times` undefined
(`NoMethodError`), since that delegating class method lives in
`active_support.rb` itself. `activesupport` depends on `tzinfo` itself, so
no separate `tzinfo`/`tzinfo-data` dependency is needed — data-source
portability (see #85's finding about `/usr/share/zoneinfo`) is the same
tradeoff #85 already documents, inherited transitively rather than added
directly.

## How

Same `:naive`-tag mechanism as #85 in `ext/duckling/src/lib.rs` (ported
unmodified, commit `7fbee40`) — Rust still only understands
`chrono::FixedOffset`, no IANA/DST concept, so the Naive/Instant
distinction has to survive the FFI boundary for Ruby to know which results
it may reinterpret.

## Test plan

- [x] Full suite green against #94's shared test suite, unmodified:
      `118 runs, 524 assertions, 0 failures, 0 errors, 6 skips`
- [x] Manually exercised the DST-transition, DST-gap/ambiguous-raises,
      zone-anchoring, zone-mismatch-raises, and `TimeWithZone`/plain-`Time`
      equality cases directly against `Duckling.parse`


cpb:
closing in favour of:

because of the more timely maintenance of the tzinfo-data gem compared to activesupport

PR #97, opened 2026-07-07
Migrate DimensionValue::Time onto the unified serde_magnus tagged shape (#91)

cpb:
## Migrate `DimensionValue::Time` onto the unified `serde_magnus` tagged shape (issue #91)

Implements #91: `:time`'s `:value` now goes through the same generic-serialize-then-patch pattern as the other 13 dimensions (from #90), instead of its own bespoke flattened shape.

**Before:**
```ruby
{ type: :value, value: 2026-07-02 00:00:00 +0000, grain: :day, values: [...] }
{ type: :interval, from: { type: :value, ... }, to: { type: :value, ... } }
```

**After:**
```ruby
{ Time: { Single: { value: { Naive: { value: 2026-07-02 00:00:00 +0000, grain: :day } },
                     values: [...], holidayBeta: "christmas" } } }
{ Time: { Interval: { from: { Naive: { ... } }, to: { Naive: { ... } }, values: [...] } } }
```

### Implementation

- `ext/duckling/src/lib.rs`: `entity_to_ruby`'s `Time` arm now calls `serialize_symbolized(ruby, &entity.value)` (same as every other dimension) to get the `Single`/`Interval` structural shape, `values`, and `holidayBeta` for free, then two new helpers patch the known-problem leaves using the typed `TimeValue`/`TimePoint` the serialization came from:
  - `patch_time_point` — overwrites one tagged `{Naive:|Instant: {value:, grain:}}` leaf: a genuine Ruby `Time` (via the existing `resolve_naive` for `Naive`, direct `IntoValue` for `Instant`) and `Grain::as_str()`'s lowercase-snake_case symbol, replacing serde's placeholder String/PascalCase output.
  - `patch_time_value` — walks to every `patch_time_point` call site: the `Single` case's primary `value` plus every `values` recurrence entry, or the `Interval` case's `from`/`to` plus each `values` recurrence entry's own `from`/`to`.
  - The old `time_value_to_ruby`/`time_point_to_ruby` hand-built-hash functions are removed; `resolve_naive` is unchanged and reused.
- Removed the now-obsolete special-case comment in the `TimeGrain`/`Duration` arms referring to Time's old bespoke path.

### Test corpus migration

Every existing assertion against the old flattened shape is rewritten onto the new tagged shape, across `duckling_test.rb` (canonical API-shape test) and the full time corpus (`duckling_parse_time_basic_test.rb`, `..._dates_test.rb`, `..._interval_test.rb`, `..._relative_test.rb`, `..._weekdays_test.rb`, `duckling_parse_locale_test.rb`, `duckling_comma_list_test.rb`). Adds a shared `time_point` helper to `test_helper.rb` for unwrapping a tagged `Naive`/`Instant` `TimePoint`, plus small per-file `single_point`/`interval_points` conveniences built on top of it.

Also corrects the hill probe test's holiday-casing assumption: verified against `duckling-0.4.0`'s own `en.rs` `"holidays"` rule (which lowercases the matched text before building `TimeForm::Holiday`), `"christmas"` resolves `holidayBeta: "christmas"`, not the Haskell/pyduckling corpus's capitalized `"Christmas"`.

### Docs

- `README.md`: rewrites the `:time` `:value` shape example and description under "Return value".
- `CHANGELOG.md`: new `[Unreleased]` breaking-change entry describing the shape change and how to migrate.
- `AGENTS.md`: updates the "Entity `:value` shapes" bullet — Time now shares the same generic-serialize-then-patch description as every other dimension, with a note on `patch_time_value`/`patch_time_point`.

### Verification

`bundle exec rake` (StandardRB + `cargo clippy -D warnings` / `cargo fmt --check` + full Minitest suite) is green: 131 runs, 597 assertions, 0 failures, 0 errors, 5 skips (pre-existing known-gap skips, unrelated to this change).

---

*Originally opened as a hill (failing-tests-first) PR; implementation, test migration, and docs landed on this same branch per the "Reviewer instructions" below.*

## Hill: failing tests for issue #91

This draft PR contains only failing tests — one commit per layer, migrated onto the parent branch via 4 merge commits (no stub commits were needed: this layer only exercises existing `Duckling.parse` behavior, no new constants/methods/routes). CI should confirm failures. **Do not merge — implementation happens on this branch.**

### time-instant-value-is-real-time
```
DucklingParseTimeBasicTest#test_now_value_is_tagged_instant_shape_with_real_time:
NoMethodError: undefined method '[]' for nil
    test/duckling_parse_time_basic_test.rb:37
```

### time-naive-resolves-reference-offset
```
DucklingParseTimeBasicTest#test_tomorrow_tagged_naive_shape:
NoMethodError: undefined method '[]' for nil
    test/duckling_parse_time_basic_test.rb:74
```

### time-interval-tagged-shape
```
DucklingParseTimeIntervalTest#test_interval_tagged_shape_3pm_to_5pm [test/duckling_parse_time_interval_test.rb:111]:
expected value to be tagged with :Time, got: {type: :interval, from: {type: :value, value: 2013-02-12 15:00:00 -0200, grain: :hour}, to: {type: :value, value: 2013-02-12 18:00:00 -0200, grain: :hour}}
```

### time-holiday-and-recurrence
```
DucklingParseTimeRelativeTest#test_christmas_holiday_and_recurrence:
NoMethodError: undefined method '[]' for nil
    test/duckling_parse_time_relative_test.rb:127
```

## Reviewer instructions

Review each expected failure message above. If they correctly specify the intended behavior per issue #91 (Time's `:value` moving onto the unified externally-tagged `serde_magnus` shape — `{Time: {Single: {value: {Naive:|Instant: {value:, grain:}}, values: [...], holidayBeta:}}}` or `{Time: {Interval: {from:, to:, values: [...]}}}`), **remove the `hill-ready` label** from this PR — that signals the implementation session to begin. Run `/continue-pr` on this PR to start it.


cpb:
## Hill ready for review

**To proceed:**
2. Open `/continue-pr 97` — that session will inspect CI and ask whether the hill is correctly specified.

test/duckling_parse_time_basic_test.rb
```diff
@@ -46,6 +69,18 @@ def test_tomorrow
     assert_equal Time.new(2013, 2, 13, 0, 0, 0, "-02:00"), entity[:value][:value]
   end
 
+  def test_tomorrow_tagged_naive_shape
+    entity = time_entity_for("tomorrow")
+    naive = entity[:value][:Time][:Single][:value][:Naive]
```
cpb, line 67: **note:** error in the hill, rather than assertion failure

test/duckling_parse_time_basic_test.rb
```diff
@@ -25,6 +25,29 @@ def test_now
...
+  # Pins the post-#90 tagged shape for :time's Single/Instant case (issue #91).
+  # The critical assertion is the explicit `is_a?(Time)` check on the datetime
+  # leaf: a generic serde_magnus serialization of DateTime<FixedOffset> would
+  # produce a plain Ruby String there, and a bare `assert_equal` against a
+  # Time literal would NOT catch that regression (String#== against a
+  # stringified Time could coincidentally match), so this test must assert
+  # the type explicitly rather than relying on equality alone.
+  def test_now_value_is_tagged_instant_shape_with_real_time
+    entity = time_entity_for("now")
+    single = entity[:value][:Time][:Single]
```
cpb, line 30: **note:** error in the hill rather than assertion failure

test/duckling_parse_time_interval_test.rb
```diff
@@ -102,6 +102,30 @@ def test_tonight_interval
     assert_equal Time.new(2013, 2, 13, 0, 0, 0, "-02:00"), value[:to][:value]
   end
 
+  def test_interval_tagged_shape_3pm_to_5pm
+    results = Duckling.parse("from 3pm to 5pm", locale: "en", reference_time: REFERENCE_TIME)
+    entity = results.find { |r| r[:dim] == :time }
+    refute_nil entity, "expected a :time entity for 'from 3pm to 5pm'"
+
+    value = entity[:value]
+    assert value.key?(:Time), "expected value to be tagged with :Time, got: #{value.inspect}"
```
cpb, line 92: **note:** failure in the hill

test/duckling_parse_time_relative_test.rb
```diff
@@ -102,4 +102,42 @@ def test_in_1_week
...
+  # 2. Recurrence: duckling returns up to 3 upcoming occurrences in `values`
+  #    (including the primary as element 0). Each element of that array must
+  #    be individually tagged the same way the top-level `value` is
+  #    (`{Naive: {value:, grain:}}` / `{Instant: {value:, grain:}}`) — the
+  #    tagging convention applies uniformly to every nested `TimePoint`, not
+  #    just the top-level one.
+  def test_christmas_holiday_and_recurrence
+    entity = time_entity("christmas")
+
+    single = entity[:value][:Time][:Single]
```
cpb, line 124: **note:** error in the hill rather than failure

test/duckling_parse_time_interval_test.rb
```diff
@@ -102,6 +102,30 @@ def test_tonight_interval
     assert_equal Time.new(2013, 2, 13, 0, 0, 0, "-02:00"), value[:to][:value]
   end
 
+  def test_interval_tagged_shape_3pm_to_5pm
+    results = Duckling.parse("from 3pm to 5pm", locale: "en", reference_time: REFERENCE_TIME)
+    entity = results.find { |r| r[:dim] == :time }
+    refute_nil entity, "expected a :time entity for 'from 3pm to 5pm'"
```
cpb: **polish (blocking):** can we refactor to use entity_for everywhere we do this kind of pattern please?

test/duckling_parse_time_relative_test.rb
cpb: **polish:** is this not just entity_for? -- lets limit proliferation of test helpers

test/duckling_parse_time_basic_test.rb
cpb: **polish:** is this not entity_for? lets limit proliferation of test helpers

cpb (commented):
take no action on the `note:` comments, action all polish across the suite.

cpb:
Addressed the blocking/polish comments: extended `entity_for` in `test_helper.rb` with an optional `reference_time:` kwarg (only forwarded to `Duckling.parse` when present — `Native.parse`'s Magnus binding treats an explicit `nil` differently from an omitted kwarg) and removed the per-file `time_entity_for`/`time_entity` duplicates in `duckling_parse_time_basic_test.rb`, `duckling_parse_time_relative_test.rb`, and `duckling_parse_time_interval_test.rb` in favor of it.

Ran the full suite locally after the change: exactly the 4 documented failures remain (`test_now_value_is_tagged_instant_shape_with_real_time`, `test_tomorrow_tagged_naive_shape`, `test_christmas_holiday_and_recurrence`, `test_interval_tagged_shape_3pm_to_5pm`), matching the PR body's "Expected failure messages" verbatim, and all other tests (including the 13 pre-existing `entity_for`-based dimension tests) still pass — so the refactor didn't change behavior anywhere outside the four intentionally-failing hill tests. Left the four "note" threads (error-vs-failure classification) unresolved since they're your own confirmations against the PR body, not action items — they check out against what I observed running the suite.

Pushed as `f3f2913`.

---

ext/duckling/src/lib.rs
```diff
@@ -421,57 +423,122 @@ fn resolve_naive(
...
-                h.aset(ruby.to_symbol("to"), time_point_to_ruby(ruby, tp, offset)?)?;
+                let point: Value = inner.aref(ruby.to_symbol("to"))?;
+                patch_time_point(ruby, point, tp, offset)?;
+            }
+
+            let vals: Value = inner.aref(ruby.to_symbol("values"))?;
+            let vals = RArray::from_value(vals).ok_or_else(|| {
+                arg_error(ruby, "expected serialized Interval values to be an Array")
+            })?;
+            for (i, endpoints) in values.iter().enumerate() {
```
cpb, line 536: **Adversarial review finding (Fable) — severity: high**

Interval recurrence `values` patching is completely untested. This `for (i, endpoints) in values.iter().enumerate()` loop (patching each `IntervalEndpoints` entry's `from`/`to`) is correct at runtime (verified: `"from 3pm to 5pm"` produces properly-patched `Interval.values` entries), but no test anywhere asserts the contents of `entity[:value][:Time][:Interval][:values]`. You could delete this entire loop and the suite would stay green — entries would silently regress to RFC3339 Strings and PascalCase grain strings (e.g. `"Hour"` instead of `:hour`). This is the most fragile code path in the migration and it ships unpinned.

Suggest adding a test asserting `interval[:values]` entries are individually tagged with real `Time` values and lowercase grain symbols, mirroring what `test_now_value_is_tagged_instant_shape_with_real_time` already does for `Single`'s recurrence.

ext/duckling/src/lib.rs
```diff
@@ -421,57 +423,122 @@ fn resolve_naive(
...
-            h.aset(ruby.to_symbol("type"), ruby.to_symbol("interval"))?;
+        TimeValue::Interval {
+            from, to, values, ..
+        } => {
+            let inner: Value = outer.aref(ruby.to_symbol("Interval"))?;
+            let inner = RHash::from_value(inner).ok_or_else(|| {
+                arg_error(ruby, "expected serialized Interval payload to be a Hash")
+            })?;
+
             if let Some(tp) = from {
```
cpb, line 523: **Adversarial review finding (Fable) — severity: medium**

Unbounded intervals now carry explicit `from: nil`/`to: nil` keys — an undocumented behavioral change with zero test coverage. On `main`, `time_value_to_ruby` only `aset` the `from`/`to` keys when `Some`, so e.g. `"after 3pm"` had *no* `:to` key at all. The new generic-serialize path always emits the key with `nil`, since `TimeValue::Interval`'s `from`/`to` fields have no `skip_serializing_if` (confirmed against `duckling-0.4.0`'s `types.rs` and at runtime: `"after 3pm"` → `{Interval: {from: {...}, to: nil, ...}}`).

Consequences:
- No test exercises the `None`-endpoint branches here (e.g. `"after 3pm"`, `"until 5pm"`).
- Neither README.md nor CHANGELOG.md mentions this — a caller migrating who previously used `value.key?(:to)` to detect an unbounded interval will now always get `true`.
- `test/duckling_test.rb`'s `test_parses_interval` assertions `interval.key?(:from)` / `interval.key?(:to)` are now tautological (they'd pass even if both were always `nil`) — see companion comment there.

Worth deciding deliberately: is the always-present-but-possibly-nil key the intended new contract, or should unbounded endpoints omit the key like before? Either way it should be documented and tested.

test/test_helper.rb
```diff
@@ -15,10 +15,24 @@
...
+
+# Issue #91: `:time`'s `:value` uses the same unified externally-tagged shape
+# as the other 13 dimensions (#90) — every `TimePoint` (a `Single` result's
+# primary `value`/each `values` entry, or an `Interval`'s `from`/`to`) is
+# individually tagged `{Naive: {value:, grain:}}` or `{Instant: {value:, grain:}}`.
+# This unwraps one tagged `TimePoint` hash down to its plain `{value:, grain:}`
+# payload regardless of which of the two tags is present, since most
+# call sites don't need to distinguish Naive from Instant.
+def time_point(tagged)
+  tagged[:Naive] || tagged[:Instant]
```
cpb: **Adversarial review finding (Fable) — severity: low**

`time_point` fails with a confusing `NoMethodError` instead of a diagnostic assertion failure on shape drift. If `tagged` is `nil` (e.g. an unbounded interval endpoint per the companion finding on `ext/duckling/src/lib.rs`'s `patch_time_value`, or an unguarded lookup like `test/duckling_comma_list_test.rb`'s `extracted_dates` reaching through `r[:value][:Time][:Single][:value]`) or carries neither `:Naive` nor `:Instant`, this returns `nil` and the caller dies with `undefined method '[]' for nil` rather than a message naming what was expected.

Suggest: `tagged[:Naive] || tagged[:Instant] || flunk("expected a :Naive- or :Instant-tagged TimePoint, got: #{tagged.inspect}")` (or similar) so failures here read as intentional shape assertions rather than crashes.

ext/duckling/src/lib.rs
```diff
@@ -421,57 +423,122 @@ fn resolve_naive(
...
+/// `Naive`, direct `IntoValue` for `Instant`) and `Grain::as_str()`'s
+/// lowercase-snake_case symbol.
+fn patch_time_point(
+    ruby: &Ruby,
+    point: Value,
+    tp: &TimePoint,
+    offset: FixedOffset,
+) -> Result<(), Error> {
+    let outer = RHash::from_value(point)
+        .ok_or_else(|| arg_error(ruby, "expected serialized TimePoint to be a Hash"))?;
```
cpb: **Adversarial review finding (Fable) — severity: low**

The `arg_error(ruby, "expected serialized ... to be a Hash/Array")` calls throughout `patch_time_point`/`patch_time_value` (this one and its siblings at ~449, 486, 494, 506, 520, 528, 564) raise `ArgumentError`, but these only fire if the serde representation drifts from the typed walk — an internal invariant violation, not anything the caller did wrong. A caller `rescue ArgumentError`-ing their own bad-`locale:`/`dims:` input handling would inadvertently swallow a real wrapper bug here. `RuntimeError` (matching the `panic_error` convention already used elsewhere in this file for "shouldn't happen" cases) would be the more apt class.

Low severity — this path is unreachable today (verified: key names all match `duckling-0.4.0`'s serde derive exactly, and every `aref` in the walk correctly errors rather than silently no-opping on a missing/renamed key), so this is a latent-defense-in-depth nit rather than an active bug.

test/duckling_test.rb
```diff
@@ -47,41 +48,44 @@ def test_parity_with_wafer_inc_duckling
...
-    assert entity[:value].key?(:from), "interval value should have :from"
-    assert entity[:value].key?(:to), "interval value should have :to"
-    assert_equal :value, entity[:value][:from][:type]
-    assert_equal :hour, entity[:value][:from][:grain]
-    assert_equal Time.new(2013, 2, 12, 15, 0, 0, "-02:00"), entity[:value][:from][:value]
-    assert_equal :value, entity[:value][:to][:type]
-    assert_equal :hour, entity[:value][:to][:grain]
+    interval = entity[:value][:Time][:Interval]
+    refute_nil interval, "Expected :value to be tagged :Interval, got: #{entity[:value].inspect}"
+    assert interval.key?(:from), "interval value should have :from"
```
cpb, line 66: **Adversarial review finding (Fable) — severity: medium (companion to the ext/duckling/src/lib.rs finding on unbounded intervals)**

`interval.key?(:from)` / `interval.key?(:to)` are now tautological. Since the new generic-serialize path always emits `from`/`to` keys (as `nil` when unbounded — `TimeValue::Interval`'s fields have no `skip_serializing_if`), these two assertions would pass even for an unbounded interval where `from`/`to` are both `nil`. This test happens to use a fully-bounded interval (`"from 3pm to 5pm"`) so it's not currently wrong, but it's not asserting what it looks like it's asserting, and there's no coverage anywhere of what an unbounded interval (e.g. `"after 3pm"`) actually produces under the new shape.

ext/duckling/src/lib.rs
cpb: Confirmed at runtime and against `duckling-0.4.0`'s `types.rs`: `TimeValue::Interval`'s `from`/`to` have no `skip_serializing_if`, so unbounded intervals now emit an explicit `to: nil` (was: key omitted entirely pre-#91). Added `test_unbounded_interval_after_3pm_has_explicit_nil_to` in `duckling_parse_time_interval_test.rb` pinning that as current behavior, but I haven't changed the Rust to omit the key — that's the "decide deliberately" part of your comment, and it changes the public contract for anyone doing `value.key?(:to)`, so I don't think it's mine to decide unilaterally. Leaving this thread open pending that call; happy to implement either direction once decided.

---

test/duckling_test.rb
cpb: Added dedicated coverage for the unbounded case (`test_unbounded_interval_after_3pm_has_explicit_nil_to` in `duckling_parse_time_interval_test.rb`) so the always-present-key behavior is at least pinned now. Left `test_parses_interval`'s `key?(:from)`/`key?(:to)` assertions here as-is since they're still accurate for this fully-bounded case — this thread and the companion `ext/duckling/src/lib.rs` one both hinge on the same open decision (keep-and-document vs. restore omit-on-None), so resolving properly depends on that.

---

test/duckling_comma_list_test.rb
```diff
@@ -44,7 +44,7 @@ class DucklingCommaListTest < Minitest::Test
   def extracted_dates(text)
     Duckling.parse(text, locale: "en", reference_time: REFERENCE_TIME)
       .select { |r| r[:dim] == :time }
-      .map { |r| r[:value][:value] }
```
cpb, line 47: **polish (if-minor):** maybe `entity_for` ?

test/test_helper.rb
```diff
@@ -15,10 +15,26 @@
...
-def entity_for(text, dim)
-  results = Duckling.parse(text, locale: "en", dims: [dim])
+# calling test if none is found. `reference_time:` is threaded through for
+# dims (like :time) whose resolution depends on an anchor moment.
+def entity_for(text, dim, reference_time: nil)
+  parse_kwargs = {locale: "en", dims: [dim.to_s]}
+  parse_kwargs[:reference_time] = reference_time if reference_time
+  results = Duckling.parse(text, **parse_kwargs)
   entity = results.find { |r| r[:dim] == dim.to_sym }
   refute_nil entity, "expected a #{dim.inspect} entity for #{text.inspect}, got: #{results.inspect}"
```
cpb, line 25: **polish:** why not flunk?

test/duckling_parse_time_interval_test.rb
```diff
@@ -5,114 +5,166 @@
...
+# Duckling.parse (issue #91's unified externally-tagged convention), not just
+# that *an* entity was found.
 class DucklingParseTimeIntervalTest < Minitest::Test
+  # Unwraps an Interval-shaped entity down to its {from:, to:} pair of plain
+  # `{value:, grain:}` payloads — see test_helper.rb's `time_point`.
+  def interval_points(entity)
+    interval = entity[:value][:Time][:Interval]
+    refute_nil interval, "Expected entity[:value][:Time] to be tagged :Interval, got: #{entity[:value].inspect}"
+    [time_point(interval[:from]), time_point(interval[:to])]
+  end
```
cpb: **polish:** extract this into test_helper like time_point, and flunk if nil

test/duckling_parse_time_dates_test.rb
```diff
@@ -17,81 +17,89 @@ def first_time_entity(text)
     entity
   end
 
+  # Unwraps a Single-shaped entity's primary tagged TimePoint down to its
+  # plain `{value:, grain:}` payload — see test_helper.rb's `time_point`.
+  def single_point(entity)
+    single = entity[:value][:Time][:Single]
+    refute_nil single, "Expected entity[:value][:Time] to be tagged :Single, got: #{entity[:value].inspect}"
+    time_point(single[:value])
+  end
```
cpb: **polish:** extract this into test_helper and flunk if nil

ext/duckling/src/lib.rs
cpb: There is no public

test/duckling_parse_time_interval_test.rb
cpb: Done — `interval_points` (and the identically-duplicated `single_point`, and `first_time_entity` in duckling_parse_time_dates_test.rb) are now in test_helper.rb, using the same `x || flunk("...")` style as `time_point`. Every per-file duplicate is removed.

---

test/duckling_parse_time_dates_test.rb
cpb: Done — extracted into test_helper.rb, and `first_time_entity` itself is gone too (it was functionally identical to `entity_for(text, :time, reference_time: REFERENCE_TIME)`, so I swapped all its call sites to that directly rather than keeping a redundant wrapper).

---

test/test_helper.rb
cpb: Left this one as `refute_nil` rather than switching to `flunk`: they're not quite equivalent here. `x || flunk(...)` only calls `flunk` — and therefore only registers as a counted Minitest assertion — when the check actually fails; on the passing path it's silent. `refute_nil(entity, msg)` runs (and counts) every time, pass or fail. `time_point`'s `tagged[:Naive] || tagged[:Instant] || flunk(...)` needed the `||` chain anyway to try both tags before giving up, so `flunk` was the natural fit there; `entity_for`'s check has no such fallback chain, so `refute_nil` stays the more direct/idiomatic choice and keeps the assertion counted. Happy to change it if you'd rather have consistency over the assertion-count difference.

---

test/duckling_comma_list_test.rb
cpb: Didn't switch this one: `entity_for` returns only the *first* matching entity, but `extracted_dates` needs *every* `:time` entity in the comma-list (that's the whole point of this file — asserting how many survive the collapsing behavior). There's no plural equivalent (`entities_for`) yet, and introducing one just for this one call site felt like more machinery than the "if-minor" ask warranted. Let me know if you'd rather have that helper added.

---

PR #102, opened 2026-07-08
Add DST-aware reference_zone: against #91's tagged Time shape

### LINKED ISSUE #96: Rebase and re-implement #85's DST-aware reference_zone: against #91's unified Time shape
## Goal

Rebase PR #85 (`reference_zone:` DST-aware Naive-grain time resolution) onto main after #91 merges, re-implementing its `:naive`-tagging and `apply_reference_zone` logic against #91's new unified `serde_magnus`-tagged `Time` `:value` shape, and update #85's description to record that the tzinfo-data vs. chrono-tz (#89) question is now settled.

## Background

PR #85 is currently draft and `CONFLICTING` against main. It adds a `:naive` boolean tag to every time-point Hash in `time_point_to_ruby` (`ext/duckling/src/lib.rs`) so that `Duckling.parse` (`lib/duckling.rb`) can tell `TimePoint::Naive` results from `TimePoint::Instant` results after conversion to a plain Ruby `Time`, and reinterpret `:naive` results' offsets against a real IANA zone via `apply_reference_zone`.

#91 migrates `DimensionValue::Time`'s `:value` onto the same externally-tagged `serde_magnus` shape used by the other 13 dimensions (`Single`/`Interval` discrimination, patched `Grain`/datetime leaves) — a structural change to the exact Hash shape #85's `:naive` tagging and `apply_reference_zone` are written against. Landing #85 as-is after #91 merges would mean patching merge conflicts against a shape the code no longer assumes; the tagging/reinterpretation logic needs to be re-derived against #91's actual shape, not just mechanically reconciled.

Separately, #85's own description left open a choice between its tzinfo/tzinfo-data approach and PR #89's alternative all-Rust `chrono-tz` implementation. That question is now settled: the alternative PR(s) have been closed in favor of tzinfo-data's runtime-patchable IANA data. #85's description should say so plainly rather than presenting it as still open.

## Acceptance criteria

- [ ] #91 has merged to main before this work starts
- [ ] PR #85 rebased onto post-#91 main
- [ ] `:naive`/`:instant` tagging (or equivalent discrimination) re-derived to work with #91's unified `Time` `:value` shape (`Single`/`Interval`, patched `Grain` and datetime leaves)
- [ ] `apply_reference_zone` (`lib/duckling.rb`) updated to walk/reinterpret the new shape correctly, including `Interval` results (`from`/`to`) and recurrence (`values`)
- [ ] All DST-transition, DST-gap/ambiguous-time-raises, zone-mismatch-raises, and unrecognized-shape-raises behavior from #85's original test plan still holds under the new shape
- [ ] `Instant`-grain arithmetic imprecision remains explicitly out of scope and documented (as in #85 today, tracked separately in #83)
- [ ] #85's PR description updated to state the tzinfo-data vs. chrono-tz question is settled (alt PR(s) closed, tzinfo-data chosen for its runtime-patchable IANA data) rather than presenting it as an open alternative
- [ ] Full suite green, `standardrb`, `cargo fmt --check`, `cargo clippy -- -D warnings` clean

## Out of scope

- Re-litigating tzinfo-data vs. chrono-tz — already decided (see above)
- `TimePoint::Instant`-grain DST arithmetic precision (#83)
- Any further changes to #91's unified shape design itself

## Notes

- Depends on #91 merging first.
- References: #85, #89 (closed alternative), #83 (Instant-grain arithmetic, out of scope), #90/#91 (shape migration this rebase targets)

cpb:
Closes #96. Supersedes #85, whose implementation predated #91's unified `serde_magnus`-tagged Time `:value` shape and could not be rebased across it; this PR re-derives the same `reference_zone:` behavior against that shape.

## Summary

`Duckling.parse` now accepts `reference_zone:`, an IANA zone name (e.g. `"America/New_York"`) that makes `Naive` (wall-clock) time results DST-aware: each result's UTC offset is resolved against the real zone for *that result's own date*, instead of the single fixed offset derived from `reference_time:` being applied uniformly. Resolution is backed by the new `tzinfo`/`tzinfo-data` runtime dependencies.

## Behavior

- `reference_zone:` + `reference_time:` → every `Naive` leaf (a `Single`'s `value`, an `Interval`'s `from`/`to`, and every generated `values` recurrence entry — all treated identically) is re-anchored in the zone, picking up its own date's offset. An interval straddling a transition gets a different offset on each leg.
- `reference_zone:` omitted → unchanged: `reference_time:`'s fixed offset applies uniformly.
- `reference_zone:` without `reference_time:` → reinterprets result offsets only; it does **not** anchor the parse, so relative expressions ("tomorrow") still anchor on the machine-local clock. Pass a `reference_time:` in the zone to anchor as well.
- `reference_time:` whose `utc_offset` disagrees with the zone's real offset at that instant → `ArgumentError` (silently preferring either would resolve results against an offset the caller never asked for).
- Unknown zone name → `ArgumentError`.
- `Instant` leaves are left untouched: their relative arithmetic already collapsed against one `FixedOffset` inside the wrapped Rust crate, and that imprecision is out of scope (#83).

### DST gaps and overlaps

A wall clock that a spring-forward gap skipped or a fall-back overlap made ambiguous resolves deterministically, matching `ActiveSupport::TimeZone#parse`/`#local` on the same input for typical zones: a gap shifts the wall clock forward by the transition's width (2:30am on a US spring-forward day becomes 3:30am EDT), an overlap takes the first (pre-transition) occurrence. This applies equally to caller-named primary values and parser-generated recurrence entries — there is no benefit to raising for one and not the other, and every `Single` carries a populated `values` array, so a raising path would be routinely reachable from ordinary input.

Two deliberate departures from ActiveSupport, both where Rails gets edge zones wrong:

- Rails hardcodes `@time += 1.hour` and retries, which overshoots `Australia/Lord_Howe`'s 30-minute gap (02:15 → 03:15 instead of 02:45); `gap_delta` reads the width off the transition itself. Every one-hour gap resolves identically under both.
- Rails resolves overlaps via tzinfo's `dst=true` flag, which only means "first occurrence" where the pre-transition period observes DST; negative-DST zones invert that (tzinfo models Europe/Dublin's winter GMT as its `dst?`-true period), so Rails returns Dublin's *second* occurrence, an hour off as an instant. `local_time_in_zone` selects the first occurrence by position instead, via the block form of `Timezone#local_time`.

## Implementation

The native extension needed no changes for this feature. #85 had to thread a `:naive` boolean through `ext/duckling/src/lib.rs` because both `TimePoint` variants collapsed into indistinguishable Ruby `Time`s; #91's shape already tags every point, so this is a pure Ruby walk (`apply_reference_zone`, `lib/duckling.rb`) over the tagged tree. The walk must stay in lockstep with `patch_time_value`/`patch_time_point` in `ext/duckling/src/lib.rs`, and raises the named `Duckling::ShapeError` on any tag it doesn't recognize so a future drift there fails loudly instead of quietly returning results resolved against the wrong offset.

The branch also carries two extension-adjacent allocation fixes found while benchmarking (see below), both in `symbolize_keys_in_place` (`ext/duckling/src/ruby_value.rs`): the `String::try_convert` type-test that built a throwaway `TypeError` (+ message + class-name String) for every already-`Symbol` key it rejected now uses the `Option`-returning `RString::from_value`, and string keys are interned via `String#to_sym` on their own VALUE instead of round-tripping through a Rust `String` plus a fresh Ruby String. Together: `medium` scenario allocations drop from 148 to ~94 objects/call, +8.4% ips on a release build.

## Dependencies

Adds `tzinfo` and `tzinfo-data` as **runtime** dependencies. `tzinfo` alone falls back to the host's zoneinfo files, which slim container images don't ship; bundling `tzinfo-data` keeps `reference_zone:` working everywhere and lets a consumer pick up a tz-database revision with a plain gem bump rather than a native recompile. Both are pure Ruby, so the precompiled-binary-gem story is unaffected. This settles the question #85 left open between the Ruby-side approach and the Rust-side alternatives (#88, #89, both closed); #85's description records the rationale.

## Benchmarks

Recorded across all five environments (`github-actions`, `claude-code-web`, `local-3.3`, `local-3.4`, `local-4.0`) — see `docs/benchmarks/`. Recording surfaced an allocation regression from #91's generic-serialize-then-patch conversion path (predating this PR): objects/call rose ~3–5× on every entity-matching scenario, identically across machines and Ruby versions. The dominant costs are fixed on this branch (see Implementation); the remainder of #91's conversion-path allocation behavior is tracked separately rather than expanded here.

## Test plan

- [x] Full suite green via `bin/test` (149 runs, 0 failures, 0 errors, 5 pre-existing skips)
- [x] `standardrb`, `cargo fmt --check`, `cargo clippy -- -D warnings` clean (CI baseline job green)
- [x] DST-transition, gap (including late-in-day gaps whose transition crosses UTC midnight — America/Nuuk), overlap (including negative-DST zones — Europe/Dublin), interval-leg, recurrence-`values`, zone-mismatch, unknown-zone, `Instant`-unaffected, and shape-drift cases each pinned by a dedicated test in `test/duckling_test.rb`


cpb:
## Hill ready for review

**To proceed:**
2. Open `/continue-pr 102` — that session will inspect CI and ask whether the hill is correctly specified.

test/duckling_reference_zone_mismatch_test.rb
```diff
@@ -0,0 +1,41 @@
...
+require_relative "test_helper"
+
+# When both `reference_time:` and `reference_zone:` are given, the fixed
+# `utc_offset` carried by `reference_time:` must agree with `reference_zone:`'s
+# real UTC offset at that instant. `reference_time:` here is built with a
+# fixed -05:00 offset, but "America/Los_Angeles" is at -07:00 (PDT) on
+# 2026-06-15 — a caller error, since there's no principled way to silently
+# prefer one over the other.
+class DucklingReferenceZoneMismatchTest < Minitest::Test
+  def test_mismatched_reference_time_offset_and_reference_zone_raises_argument_error
```
cpb: **polish:** move into duckling_test

test/duckling_reference_zone_unrecognized_shape_test.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative "test_helper"
+
+# Characterizes apply_reference_zone's loud-failure behavior when a :time
+# entity's :value doesn't match the known Single/Interval shape — a future
+# drift in the Rust side's serialized shape must not pass unnoticed.
+class DucklingReferenceZoneUnrecognizedShapeTest < Minitest::Test
+  def test_raises_on_unrecognized_time_value_shape
```
cpb: **issue:** move into duckling_test

test/duckling_reference_zone_interval_test.rb
```diff
@@ -0,0 +1,40 @@
...
+# an Interval-shaped time result's `from` and `to` legs must each be
+# reinterpreted against `reference_zone:` INDEPENDENTLY, using each leg's own
+# date's real UTC offset — not a single offset borrowed from reference_time:
+# or from just one of the two legs. "from March 7th 2026 3:00am to March 9th
+# 2026 3:00am" straddles the US spring-forward transition at 2:00am local on
+# 2026-03-08: the `from` leg (March 7, still standard time) must resolve to
+# EST (UTC-5) while the `to` leg (March 9, already daylight time) must
+# resolve to EDT (UTC-4).
+class DucklingReferenceZoneIntervalTest < Minitest::Test
+  def test_interval_legs_resolve_independently_across_dst_transition
```
cpb: **issue:** move into duckling_test

test/duckling_reference_zone_instant_test.rb
```diff
@@ -0,0 +1,42 @@
...
+
+# Issue #85 (DST-aware reference_time via reference_zone:): TimePoint::Instant
+# results (e.g. "in 3 hours" — relative/duration-based, already resolved to an
+# absolute instant by the underlying Rust crate against a single FixedOffset
+# before this gem ever sees it) must NOT be reinterpreted by reference_zone:.
+# That arithmetic imprecision is explicitly out of scope (tracked separately
+# in issue #83) — reference_zone:'s DST awareness only applies to :Naive
+# (wall-clock) results.
+class DucklingReferenceZoneInstantTest < Minitest::Test
+  def test_reference_zone_leaves_instant_result_unaffected
```
cpb: **issue:** move into duckling_test

test/duckling_reference_zone_dst_transition_test.rb
```diff
@@ -0,0 +1,46 @@
...
+# (wall-clock) time result DST-aware by resolving its UTC offset against the
+# real IANA zone for *that result's own date*, instead of the single fixed
+# offset `reference_time:` provides today. US DST began 2026-03-08 02:00
+# local (America/New_York): "spring forward" from EST (UTC-5) to EDT (UTC-4).
+# A naive result dated just before that transition must resolve to -18000
+# (EST); one dated just after must resolve to -14400 (EDT) — proving each
+# result is resolved against its own date, not a single offset applied
+# uniformly across both.
+class DucklingReferenceZoneDstTransitionTest < Minitest::Test
+  def test_naive_results_resolve_offset_per_date_across_dst_transition
```
cpb: **issue:** move into duckling_test

test/duckling_reference_zone_dst_gap_test.rb
```diff
@@ -0,0 +1,26 @@
...
+
+# reference_zone: (new keyword, not yet supported by Duckling.parse) makes
+# :Naive wall-clock time results DST-aware by resolving each result's UTC
+# offset against the real IANA zone for that result's own date, instead of
+# blindly inheriting reference_time's fixed offset. America/New_York springs
+# forward from 2:00am straight to 3:00am on 2026-03-08, so "2:30am" that day
+# is a local time that never actually occurs — it must raise rather than
+# silently pick an arbitrary (wrong) offset.
+class DucklingReferenceZoneDstGapTest < Minitest::Test
+  def test_raises_argument_error_for_dst_spring_forward_gap
```
cpb: **issue:** move into duckling_test

cpb:
Reviewed. CI failed with exactly the 6 documented assertion failures (no load errors) — confirmed against the job logs. Checked each new test file against issue #96's acceptance criteria and the unified Time `:value` shape (`Single`/`Interval`, `Naive`/`Instant`-tagged `TimePoint` leaves):

- **dst-transition** / **interval-endpoints**: assert on `.utc_offset` per-date across the 2026-03-08 America/New_York spring-forward, correctly indexed through `[:value][:Time][:Single][:value][:Naive][:value]` / `[:value][:Time][:Interval][:from|:to][:Naive][:value]`.
- **dst-gap-ambiguous**: asserts `ArgumentError` for a wall-clock time that falls in the nonexistent 2:00–3:00am spring-forward gap.
- **zone-mismatch**: asserts `ArgumentError` with an "offset" message, and explicitly guards against being satisfied by the unrelated "unknown keyword: :reference_zone" error the current stub would otherwise raise.
- **instant-unaffected**: asserts `TimePoint::Instant` results pass through unchanged and that an unrecognized zone name raises `ArgumentError`.
- **unrecognized-shape**: asserts `apply_reference_zone` raises `RuntimeError` on a malformed `:time` value shape.

All correctly specify the intended behavior. Removing the `hill-ready` label to signal the implementation session to begin.

---

test/duckling_reference_zone_dst_transition_test.rb
```diff
@@ -0,0 +1,46 @@
...
+      reference_time: REFERENCE_TIME,
+      reference_zone: "America/New_York"
+    )
+    after_results = Duckling.parse(
+      "March 9th 2026 3:00am",
+      locale: "en",
+      dims: ["time"],
+      reference_time: REFERENCE_TIME,
+      reference_zone: "America/New_York"
+    )
```
cpb: **polish:** use the test helpers?

test/duckling_reference_zone_instant_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    without_zone = entity_for("in 3 hours", :time, reference_time: REFERENCE_TIME)
+    with_zone = Duckling.parse(
+      "in 3 hours",
+      locale: "en",
+      dims: ["time"],
+      reference_time: REFERENCE_TIME,
+      reference_zone: "America/New_York"
+    ).find { |r| r[:dim] == :time }
```
cpb: **polish:** use the test helpers (add optional reference_zone)

test/duckling_reference_zone_interval_test.rb
```diff
@@ -0,0 +1,40 @@
...
+    results = Duckling.parse(
+      "from March 7th 2026 3:00am to March 9th 2026 3:00am",
+      locale: "en",
+      dims: ["time"],
+      reference_time: REFERENCE_TIME,
+      reference_zone: "America/New_York"
+    )
+    entity = results.find { |r| r[:dim] == :time }
+    refute_nil entity, "expected a :time entity, got: #{results.inspect}"
```
cpb: **polish:** use the test helpers

lib/duckling.rb
```diff
@@ -39,11 +39,21 @@ def self.parse(*args, **kwargs, &block)
       kwargs = kwargs.merge(reference_time: reference_time.to_time)
     end
 
-    return Native.parse(*args, **kwargs, &block) unless Fiber.scheduler
+    reference_zone = kwargs.delete(:reference_zone)
```
cpb: **Critical — branch is missing main's PR #101 (stale rebase)**

`Duckling.parse`'s signature here (line 36) is still `def self.parse(*args, **kwargs, &block)`, but `origin/main` (commit `492ca70`, "Give Duckling.parse an explicit Ruby signature", landed after this branch's merge-base `bef4e24`) already has the explicit-keyword signature: `def self.parse(text, locale: "en", dims: ["time"], reference_time: nil, with_latent: false)`.

Concretely: `Duckling.parse("tomorrow", reference_time: nil)` succeeds on `main` but raises `TypeError: no implicit conversion of NilClass into Time` on this branch, because the old `*args, **kwargs` body forwards an explicitly-passed `nil` straight to `Native.parse` instead of dropping it (main's version does `kwargs[:reference_time] = reference_time if reference_time`).

Rebase this branch onto current `main` and re-derive the `reference_zone:`/`kwargs.delete` stub (this line) against #101's explicit signature.

*(fable, adversarial review)*

test/duckling_test.rb
```diff
@@ -104,4 +104,151 @@ def test_date_time_reference_time_is_coerced_and_preserves_utc_offset
...
+      "got #{to_point[:value].utc_offset} (#{to_point[:value].inspect})")
+  end
+
+  # When both `reference_time:` and `reference_zone:` are given, the fixed
+  # `utc_offset` carried by `reference_time:` must agree with
+  # `reference_zone:`'s real UTC offset at that instant. `reference_time:`
+  # here is built with a fixed -05:00 offset, but "America/Los_Angeles" is at
+  # -07:00 (PDT) on 2026-06-15 — a caller error, since there's no principled
+  # way to silently prefer one over the other.
+  def test_reference_zone_mismatched_reference_time_offset_raises_argument_error
```
cpb, line 251: **Critical — test suite is internally unsatisfiable**

This test establishes the contract that `Duckling.parse` must raise `ArgumentError` whenever `reference_time:`'s fixed offset disagrees with `reference_zone:`'s real UTC offset at that instant (per the comment above: "there's no principled way to silently prefer one over the other").

But three other tests in this file pass `reference_time: REFERENCE_TIME` (`Time.new(2013, 2, 12, 4, 30, 0, "-02:00")`, from `test/test_helper.rb`) together with `reference_zone: "America/New_York"`, and expect *normal, non-raising* DST-aware resolution:
- `test_reference_zone_resolves_naive_offset_per_date_across_dst_transition` (line 117)
- `test_reference_zone_leaves_instant_result_unaffected` (line 159)
- `test_reference_zone_resolves_interval_legs_independently_across_dst_transition` (line 192)

America/New_York's real UTC offset at REFERENCE_TIME's instant (2013-02-12 06:30 UTC) is `-18000` (EST), not `-7200` — verified directly:
```
ruby -e 'ENV["TZ"]="America/New_York"; puts Time.at(Time.new(2013,2,12,4,30,0,"-02:00").to_i).localtime.utc_offset'
# => -18000
```
So by this test's own contract, those three tests should *also* raise `ArgumentError`, directly contradicting their own non-raising assertions — no single implementation can satisfy all four as written. Either give the three DST tests a `reference_time:` whose offset actually agrees with `America/New_York` (the way `test_reference_zone_raises_argument_error_for_dst_spring_forward_gap`, line 138, already does — `Time.new(2026, 3, 1, 9, 0, 0, "-05:00")`), or explicitly narrow this test's contract.

*(fable, adversarial review)*

test/duckling_test.rb
```diff
@@ -104,4 +104,151 @@ def test_date_time_reference_time_is_coerced_and_preserves_utc_offset
...
+  # results.
+  def test_reference_zone_leaves_instant_result_unaffected
+    without_zone = entity_for("in 3 hours", :time, reference_time: REFERENCE_TIME)
+    with_zone = entity_for("in 3 hours", :time,
+      reference_time: REFERENCE_TIME, reference_zone: "America/New_York")
+
+    instant_without = single_point(without_zone)[:value]
+    instant_with = single_point(with_zone)[:value]
+
+    assert_equal instant_without, instant_with,
```
cpb, line 191: **Medium — assertion can't detect the bug it's meant to catch**

\`assert_equal instant_without, instant_with\` relies on \`Time#==\`, which compares only the instant (\`to_r\`), not \`utc_offset\`:
\`\`\`ruby
a = Time.new(2013,1,1,0,0,0,"-02:00")
b = a.getlocal("-05:00")
a == b # => true, despite different utc_offset
\`\`\`
An implementation that wrongly re-zones an \`Instant\` result's displayed offset while preserving the same instant (e.g. a \`.getlocal("America/New_York")\`-style bug) would pass this assertion even though it violates the "completely unaffected" claim in the comment above. This file already knows the pattern to avoid this — lines 79-89 explicitly assert \`.utc_offset\` for exactly this reason. Suggest adding \`assert_equal instant_without.utc_offset, instant_with.utc_offset\` here too.

*(fable, adversarial review)*

test/duckling_test.rb
```diff
@@ -104,4 +104,151 @@ def test_date_time_reference_time_is_coerced_and_preserves_utc_offset
...
+  def test_reference_zone_resolves_naive_offset_per_date_across_dst_transition
+    before_entity = entity_for("March 7th 2026 3:00am", :time,
+      reference_time: REFERENCE_TIME, reference_zone: "America/New_York")
+    after_entity = entity_for("March 9th 2026 3:00am", :time,
+      reference_time: REFERENCE_TIME, reference_zone: "America/New_York")
+
+    before_value = single_point(before_entity)[:value]
+    after_value = single_point(after_entity)[:value]
+
+    assert_equal(-18000, before_value.utc_offset,
```
cpb, line 141: **Medium — offset-only assertions don't pin the wall-clock time**

This test (and \`test_reference_zone_resolves_interval_legs_independently_across_dst_transition\`, lines 197/202) asserts only \`.utc_offset\` on the resolved \`Time\` values, never the wall-clock hour or a full expected \`Time\`. Verified that a wrong, instant-preserving zone conversion satisfies both offset assertions while silently losing the "3:00am" wall clock:
\`\`\`ruby
Time.new(2026,3,7,3,0,0,"-02:00").getlocal("America/New_York") # offset -18000, hour 0 — not 3
Time.new(2026,3,9,3,0,0,"-02:00").getlocal("America/New_York") # offset -14400, hour 1 — not 3
\`\`\`
Suggest adding an assertion on \`.hour\` (or comparing against the full expected \`Time\`, e.g. \`Time.new(2026,3,7,3,0,0,"-05:00")\`) in both tests so a \`getlocal\`-style bug can't pass.

*(fable, adversarial review)*

test/duckling_test.rb
```diff
@@ -104,4 +104,151 @@ def test_date_time_reference_time_is_coerced_and_preserves_utc_offset
...
+    instant_without = single_point(without_zone)[:value]
+    instant_with = single_point(with_zone)[:value]
+
+    assert_equal instant_without, instant_with,
+      "expected reference_zone: to leave a TimePoint::Instant result's resolved Time completely unaffected"
+
+    # reference_zone: must be validated as a real IANA zone name — an
+    # unrecognized identifier should be rejected rather than silently
+    # accepted and ignored.
+    assert_raises(ArgumentError) do
```
cpb, line 202: **Low — unrelated behavior bundled into one test**

The zone-name-validation check (lines 170-181: \`Duckling.parse(..., reference_zone: "Not/A/Real/Zone")\` must raise \`ArgumentError\`) is bundled into \`test_reference_zone_leaves_instant_result_unaffected\`, a test whose name and primary assertion are about \`Instant\` results being unaffected by \`reference_zone:\`. If the instant-unaffected assertion fails first, the invalid-zone-name check never runs, and a failure here reports under a misleading test name. Consider splitting this into its own \`test_reference_zone_rejects_unrecognized_zone_name\`.

*(fable, adversarial review)*

test/duckling_test.rb
```diff
@@ -104,4 +104,151 @@ def test_date_time_reference_time_is_coerced_and_preserves_utc_offset
...
+  end
+
+  # America/New_York springs forward from 2:00am straight to 3:00am on
+  # 2026-03-08, so "2:30am" that day is a local time that never actually
+  # occurs — it must raise rather than silently pick an arbitrary (wrong)
+  # offset.
+  def test_reference_zone_raises_argument_error_for_dst_spring_forward_gap
+    reference_time = Time.new(2026, 3, 1, 9, 0, 0, "-05:00")
+
+    assert_raises(ArgumentError) do
```
cpb: **Low — inconsistent with this file's own impostor-guard pattern**

This test's bare \`assert_raises(ArgumentError)\` doesn't check the error message. \`test_reference_zone_mismatched_reference_time_offset_raises_argument_error\` (line 214) establishes the pattern of asserting on the message specifically to rule out an unrelated \`ArgumentError\` impostor satisfying the assertion by accident (see its comment, lines 225-233). Since \`reference_zone:\` currently gets deleted from kwargs before reaching \`Native.parse\` (\`lib/duckling.rb:42\`), any other unrelated \`ArgumentError\` in the call path would currently satisfy this bare \`assert_raises\` too. Consider asserting the message mentions the nonexistent local time / DST gap, for the same reason the sibling test does.

*(fable, adversarial review)*

cpb:
**Fable adversarial review (posted via Sonnet) — summary**

Ran an 8-angle adversarial review (correctness ×3, cleanup ×3, altitude, conventions) at high effort, verified each surviving candidate empirically. 6 findings posted inline above (2 critical, 2 medium, 2 low). One more doesn't anchor to a diff line since it's about an unchanged file:

**Low — AGENTS.md is now out of date with this diff (per its own "Keeping this file current" rule)**

AGENTS.md's "Public and internal APIs" section documents `Duckling.parse(text, locale: "en", dims: ["time"], reference_time: nil, with_latent: false)` and states reference_time coercion "is the one real behavioral difference between the two entrypoints." After this diff, `Duckling.parse` also silently accepts (and strips) `reference_zone:`, which `Native.parse` does not — a second, undocumented divergence. AGENTS.md itself says: "If you're an agent and notice this file is out of date with what you just observed in the repo, fix it in the same PR rather than working around the discrepancy silently." Since this is a hill-first PR and `reference_zone:`'s real semantics aren't implemented yet, it's reasonable to defer this update to the implementation PR — flagging so it isn't forgotten.

The two critical findings are the ones most worth resolving before this PR's `hill-ready` review proceeds: the branch is missing main's PR #101 (stale rebase, real behavioral regression), and the DST-transition tests are mutually contradictory with the zone-mismatch test as currently written (no single implementation can satisfy both).

cpb:
Addressed the adversarial review findings:

**Critical — stale rebase (#101 missing)**: rebased onto current `main` (`492ca70`), re-deriving the `reference_zone:`/`apply_reference_zone` stub against #101's explicit `Duckling.parse` signature. Verified `Duckling.parse("tomorrow", reference_time: nil)` no longer raises `TypeError`.

**Critical — internally unsatisfiable test suite**: gave the three DST tests (`..._resolves_naive_offset_per_date_across_dst_transition`, `..._leaves_instant_result_unaffected`, `..._resolves_interval_legs_independently_across_dst_transition`) a `reference_time:` (`Time.new(2026, 3, 1, 9, 0, 0, "-05:00")`) that itself agrees with America/New_York's real offset, so they no longer contradict the offset-mismatch test's contract.

**Medium × 2**: added `.hour` assertions alongside `.utc_offset` in the DST-transition/interval tests, and a `.utc_offset` assertion for the instant-unaffected test — both close the `getlocal`-style-bug gaps you identified.

**Low × 2**: split the invalid-zone-name check into its own `test_reference_zone_rejects_unrecognized_zone_name`, and pinned the DST-gap test's `ArgumentError` message (ruling out the same "unknown keyword" impostor the zone-mismatch test already guards against).

Ran the full suite locally — all 6 hill failures still reproduce with the expected messages (confirmed unchanged: the split-out zone-name test now carries that specific failure independently). StandardRB, `cargo fmt --check`, and `cargo clippy -- -D warnings` are clean. Pushed as `ee8d76b` (force-pushed due to the rebase).

The AGENTS.md documentation-staleness note is acknowledged — deferring it to the implementation PR as suggested, since `reference_zone:`'s real semantics aren't implemented yet on this hill branch.

---

lib/duckling.rb
```diff
@@ -33,17 +35,132 @@ module Duckling
...
+
+  # The Rust side already resolved this Naive wall-clock against reference_time:'s
+  # fixed offset, so the Time's own calendar fields still read as that intended
+  # wall-clock — re-anchoring those same fields in `zone` is what picks up the
+  # per-date DST offset.
+  #
+  # A wall-clock that a spring-forward gap skipped, or that a fall-back overlap
+  # made ambiguous, has no single correct offset; surface that as an ArgumentError
+  # about the caller's input rather than leaking a TZInfo internal.
+  def self.local_time_in_zone(zone, time)
```
cpb, line 186: **Finding 1 (highest severity): a valid query can be killed by a DST transition it never asked about.**

This guard is applied to *every* `Naive` leaf, including each entry of a `Single`'s `values` recurrence array. Those entries are future occurrences the parser generated, not times the caller named. If any one lands in a gap or an overlap, the whole `parse` call dies. Reproduced against this branch:

```ruby
Duckling.parse("every sunday at 2:30am",
  reference_time: Time.new(2026, 2, 28, 9, 0, 0, "-05:00"),
  reference_zone: "America/New_York")

# primary value would be 2026-03-01 02:30 (perfectly valid)
# values = [03-01, 03-08]
# => ArgumentError: 2026-03-08 02:30:00 does not exist in America/New_York
#    (skipped by a daylight-saving spring-forward transition)
```

The same query without `reference_zone:` returns cleanly. The fall-back side behaves identically — `"every sunday at 1:30am"` anchored in late October 2026 raises `ArgumentError: 2026-11-01 01:30:00 is ambiguous`.

Raising for the **primary** `value` is defensible: if the caller literally says "March 8th 2:30am" in New York, that wall clock does not exist and there is no honest answer. But a **recurrence entry** is collateral — one bad occurrence out of many should not destroy the whole result.

For `values` entries, pick one of: skip the entry, or resolve it deterministically (TZInfo's `local_time` takes a `dst` disambiguation argument, and `period_for_local(..., false)` lets you take the first of an ambiguous pair). Whichever you choose, state it in the comment — `AGENTS.md` currently documents "a wall-clock skipped by a gap raises" without disclosing that a date the caller never mentioned can trigger it.

lib/duckling.rb
```diff
@@ -33,17 +35,132 @@ module Duckling
...
+      next unless entity[:dim] == :time
+      reinterpret_time_value!(entity[:value][:Time], zone)
+    end
+    entities
+  end
+
+  def self.reinterpret_time_value!(value, zone)
+    if (single = value && value[:Single])
+      reinterpret_time_point!(single[:value], zone)
+      single[:values]&.each { |point| reinterpret_time_point!(point, zone) }
```
cpb, line 125: **Finding 2: this walk is untested, and it is not an exotic path.**

The handoff frames `values` as a known gap that only matters for recurrences. It's more load-bearing than that: **every** `Single` result carries a populated `values` array. `"monday at 3am"` — no recurrence word anywhere — produces two entries. So this line runs on essentially every time parse, with zero assertions covering it.

The good news is it's correct. Verified directly rather than by hand-check:

```ruby
Duckling.parse("every monday at 3am",
  reference_time: Time.new(2026, 3, 1, 9, 0, 0, "-05:00"),
  reference_zone: "America/New_York")
# values => [2026-03-02 03:00:00 -0500, 2026-03-09 03:00:00 -0400]
```

Exactly right — same wall clock, EST before the transition and EDT after.

Two tests should land before this leaves draft:
1. A `Single`'s `values` asserting that `-0500` / `-0400` pair across the transition.
2. An `Interval`'s `values` endpoint pair asserting `from`/`to` resolve independently.

Both are cheap, and both would catch a regression the six existing tests sail straight past.

lib/duckling.rb
```diff
@@ -33,17 +35,132 @@ module Duckling
...
+  end
+  private_class_method :reinterpret_interval_endpoints!
+
+  def self.reinterpret_time_point!(point, zone)
+    return if point.nil?
+
+    if (naive = point[:Naive])
+      naive[:value] = local_time_in_zone(zone, naive[:value])
+    elsif !point.key?(:Instant)
+      raise "unrecognized TimePoint shape, expected a :Naive or :Instant tag: #{point.inspect}"
```
cpb: **`RuntimeError` here is the right class** (see the review body). One improvement: give it a named subclass, e.g. `Duckling::ShapeError < RuntimeError`.

Two reasons. It's greppable, and it stops `assert_raises(RuntimeError)` from being satisfiable by an unrelated `RuntimeError` — notably the native panic path, which also raises `RuntimeError` (`panic_error` in `lib.rs`).

Also: the `TimeValue`-layer raise has a test (`test_reference_zone_raises_on_unrecognized_time_value_shape`), but **this `TimePoint`-layer raise does not.** Worth one more malformed-shape case.

lib/duckling.rb
```diff
@@ -33,17 +35,132 @@ module Duckling
...
+
+    zone = timezone_for(reference_zone)
+    entities.each do |entity|
+      next unless entity[:dim] == :time
+      reinterpret_time_value!(entity[:value][:Time], zone)
+    end
+    entities
+  end
+
+  def self.reinterpret_time_value!(value, zone)
```
cpb, line 122: **Sharp edge, undocumented and untested:** `reference_zone:` given *without* `reference_time:` silently anchors relative expressions on machine-local "now", not on "now in that zone". The zone only reinterprets results after the fact, and `verify_reference_time_offset!` is skipped entirely.

So `Duckling.parse("tomorrow", reference_zone: "Asia/Tokyo")` on a US host can resolve to the wrong calendar day. That deserves at least a line in the comment block above `parse`, if not an explicit test.

test/duckling_test.rb
```diff
@@ -104,4 +104,189 @@ def test_date_time_reference_time_is_coerced_and_preserves_utc_offset
...
+    # own unrelated ArgumentError ("unknown keyword: :reference_zone") —
+    # itself an ArgumentError, so a bare `assert_raises(ArgumentError)` would
+    # be satisfied by that impostor without any real offset-mismatch check
+    # ever running. Assert on the message to rule that impostor out and pin
+    # down the actual semantic complaint we want.
+    refute_match(/unknown keyword/i, error.message,
+      "expected the offset-mismatch ArgumentError, but got the unrelated " \
+      "'unknown keyword' error raised because reference_zone: isn't " \
+      "recognized/validated yet: #{error.message.inspect}")
+    assert_match(/offset/i, error.message,
```
cpb: The impostor guard holds, but by less margin than it looks. `refute_match(/unknown keyword/i)` plus `assert_match(/offset/i)` does exclude Magnus's keyword error — but the Rust side's `resolve_naive` failure is **also** an `ArgumentError` and **also** contains the word "offset":

> `invalid or ambiguous naive time for reference offset`

Tighten the positive assertion to pin the complaint you actually mean, e.g. `assert_match(/does not match reference_zone/, error.message)`.

duckling.gemspec
```diff
@@ -43,6 +43,17 @@ Gem::Specification.new do |spec|
   # close to whatever's actually locked in Gemfile.lock.
   spec.add_dependency "rb_sys", "~> 0.9.128"
 
+  # Resolves reference_zone:'s per-date IANA offsets (lib/duckling.rb). The
+  # wrapped Rust crate only knows a single FixedOffset, so DST-aware
+  # resolution needs a real tz database on the Ruby side.
+  spec.add_dependency "tzinfo", "~> 2.0"
```
cpb, line 49: Sanity-checked against the precompiled-binary-gem story in `AGENTS.md`'s "Build model" — **no problem.** `tzinfo-data` is pure Ruby and platform-independent, so a binary gem consumer just installs one more pure-Ruby gem; it doesn't touch the cross-compilation matrix at all.

`~> 1.2024` is also correct as written: two-component pessimistic constraint, so it allows `1.2025.x` and later data releases (which is exactly the freshness you want) while excluding a `2.0` major bump.

cpb (commented):
Adversarial review of the landed implementation. **Not a merge signal — this PR stays draft.**

Two things should gate leaving draft: the `values` walk needs tests (it runs on essentially every parse, not just explicit recurrences), and the recurrence-gap raise needs a decision written down either way. Everything else is polish.

### Verified as correct

- **The extension really was untouched, and it really didn't need touching.** I checked `patch_time_value`/`patch_time_point` in `ext/duckling/src/lib.rs` rather than taking the claim on faith. The `Single`/`Interval` and `Naive`/`Instant` shapes the Ruby walk assumes match exactly, including that an `Interval` leg is `Option<TimePoint>` serialized as a present key holding `nil`.
- **`local_time_in_zone` is not a `getlocal`-style bug.** Wall clock preserved, offset recomputed — confirmed empirically (`3am` stays `3am`, offset moves `-0500` → `-0400`). The `.hour` assertions would catch the bug: a `getlocal` shift turns `2026-03-09 03:00 -0500` into `02:00 -0400`, failing `assert_equal 3, after_value.hour`.
- **`Instant` leaves are genuinely untouched**, including legs nested in an `Interval`.
- **Raising on the offset mismatch is right.** Keep #85's choice. Zone-wins silently discards an offset the caller built on purpose; fixed-offset-wins makes `reference_zone:` a lie.
- **`RuntimeError` for shape drift is a correct divergence from #85.** It matches `internal_error()` in `lib.rs`, which already uses `RuntimeError` for exactly this "serde drifted from the typed walk" class, with the same rationale: a caller rescuing `ArgumentError` around bad `locale:`/`reference_zone:` input must not swallow a wrapper bug.

See inline comments for the findings.

cpb:
Addressed the review in `ffaf10d`.

**Finding 1 — recurrence-gap decision: resolve deterministically.** The DST guard no longer runs uniformly on every `Naive` leaf. Primary values still raise (a wall-clock the caller literally named that doesn't exist has no honest answer), but generated `values` entries resolve deterministically instead, keeping the wall-clock hour:
- **spring-forward gap** → takes the transition's post-transition offset (e.g. `2026-03-08 02:30` in a recurrence → `02:30 -0400`)
- **fall-back overlap** → takes the first (pre-transition) occurrence (`dst: true`)

So `parse("every sunday at 2:30am", reference_zone: "America/New_York", …)` now returns cleanly instead of dying on one collateral occurrence. Written down in the code comment on `local_time_in_zone` and in AGENTS.md's `reference_zone:` bullet.

**Finding 2 — values walk now tested.** Added a `Single` `values` pair (`-0500`/`-0400` across the transition) and an `Interval` `values` endpoint pair (`from`/`to` resolving independently), plus the two new deterministic gap/overlap recurrence cases.

**Other findings:**
- Named the shape-drift error `Duckling::ShapeError < RuntimeError` (greppable; can't be satisfied by the native-panic `RuntimeError`) and added the missing `TimePoint`-layer malformed-shape test.
- Tightened the offset-mismatch assertion to `/does not match reference_zone/` so the Rust-side `resolve_naive` `ArgumentError` (also contains "offset") can't impersonate it.
- Documented the `reference_zone:`-without-`reference_time:` anchoring edge.
- gemspec `~> 1.2024` finding needed no change (agreed).

Full suite green locally (145 runs, 0 failures). Still draft — not a merge request.

---

PR #54, opened 2026-07-02
Extend Ruby time-test corpus; return real Time objects for time values

cpb:
## Summary

Implements the extended Ruby time-test corpus from `ruby-test-design.md`
(issue #34) — 7 test classes covering basic relative time, weekdays, absolute
dates, relative expressions, intervals, latent entities, and locale handling.

While rebasing this hill onto `main` after #45 landed, the reviewed
direction expanded: `Duckling.parse` now returns **real Ruby `Time` objects**
for time values, for both `Naive` (wall-clock, e.g. "tomorrow", "5pm") and
`Instant` (e.g. "in one hour") grains, instead of formatted ISO-ish strings.
`reference_time:`'s offset is threaded through `Context::timezone()` and
applied to `Naive` values at serialization time — which is exactly what
issue #33 asked for (Naive values resolved against the reference zone,
ActiveSupport-style), just delivered as a `Time` object rather than an
offset-aware string. The outer Hash shape (`:type`/`:grain`/`:values`/
`:from`/`:to` keys) is unchanged — only what `:value` *is* changes, so this
doesn't encroach on #32's separate Data-object-shape scope.

Closes #34
Closes #33

## What changed

- **`ext/duckling/Cargo.toml`**: added magnus's `chrono` feature — this is
  what makes `chrono::DateTime<FixedOffset>`/`NaiveDateTime` implement
  magnus's `IntoValue`, so a parsed time value can be handed to Ruby as a
  real `Time` instead of a formatted string.
- **`ext/duckling/src/lib.rs`**: threads the reference offset
  (`Context::timezone()`) through `entity_to_ruby` → `time_value_to_ruby` →
  `time_point_to_ruby`; resolves `TimePoint::Naive` against it via a shared
  `resolve_naive()` helper, `TimePoint::Instant` passes through directly
  (already offset-aware).
- **`test/duckling_time_test.rb`**: the 7 corpus classes from
  `ruby-test-design.md`, all assertions comparing against `Time` objects
  (not strings) — including explicit `.utc_offset` checks wherever a test's
  whole point is verifying offset preservation, since `Time#==` alone only
  compares the instant.
- **`test/duckling_test.rb`**, **`test/duckling_comma_list_test.rb`**:
  existing `:value`/`:from`/`:to` assertions updated to the new `Time`
  contract.
- **`README.md`**, **`CHANGELOG.md`**: documented as a further breaking
  change on top of #45's `reference_time:` change — `:value` is now `Time`,
  not `String`, for both grains.
[...]

PR #113, opened 2026-08-04
Add two platforms, correct the precompiled gems, and test them before release

### LINKED ISSUE #112: Precompiled gems never reached RubyGems: cut a release and extend the platform matrix
The cross-compile pipeline added in #60 (closing #43) is wired end to end, but **no precompiled gem has ever been published**. Every version on RubyGems is source-only, so every consumer compiles the Rust extension at `bundle install` and needs a full cargo/rustc toolchain on the build machine.

```
$ curl -s https://rubygems.org/api/v1/versions/duckling.json | jq -r '.[] | "\(.number) | \(.platform)"'
0.2.0 | ruby
0.1.2 | ruby
0.1.1 | ruby
0.1.0 | ruby
```

The GitHub release assets agree — `v0.2.0` carries only `duckling-0.2.0.gem`.

## Why

A timing gap, not a bug in the workflow:

| | |
|---|---|
| `v0.2.0` release run | 2026-07-01 04:57:53Z |
| `cross-gem.yml` merged (#60) | 2026-07-02 18:13:49Z |

The cross-compile job landed ~37 hours *after* the most recent release, so `release.yml` has never actually executed it. The plumbing looks correct — `publish` declares `needs: [ci, cross_gems]`, downloads the `native-gem-*` artifacts into `pkg/`, and loops `gem push` over `pkg/*.gem` — it just hasn't run yet.

## What's missing

**1. A release that exercises the pipeline.** Until a new tag is cut, the workflow is unverified: this would be its first real run, so it may well surface issues (artifact naming, `merge-multiple`, RubyGems accepting multiple platforms under one version).

**2. Platform coverage.** The current matrix is `[x86_64-linux, x86_64-darwin]`. Missing:

- **`arm64-darwin`** — Apple Silicon. Every developer on an M-series Mac compiles from source today.
- **`aarch64-linux`** — arm64 Linux: arm CI runners, Docker on Apple Silicon, arm64 cloud instances.
- **`x86_64-linux-musl`** — Alpine containers. Lower priority.

`ext/duckling/Cargo.toml` depends only on `magnus`, `duckling`, `chrono`, and `rb-sys` — all pure Rust with no C library linkage — so these targets should be reachable through `rb-sys-dock` without new system dependencies. The existing `ruby-versions: "3.2"` choice carries over unchanged: `stable-api-compiled-fallback` means one binary per platform still covers every Ruby >= 3.2.

## Consumer impact

Adopting the gem in a Rails app deployed `x86_64-linux` fails at build time without a Rust toolchain. Adding a Rust toolchain just to install a single gem is a poor trade.

For contrast, the other rb-sys gems in typical apps resolve cleanly because they publish native gems — this is what a working entry looks like in `Gemfile.lock`:

```
tiktoken_ruby (0.0.9)
  rb_sys (= 0.9.87)
tiktoken_ruby (0.0.9-aarch64-linux)
tiktoken_ruby (0.0.9-arm64-darwin)
tiktoken_ruby (0.0.9-x86_64-darwin)
tiktoken_ruby (0.0.9-x86_64-linux)

duckling (0.2.0)          <- no platform variants to resolve
  rb_sys (~> 0.9.39)
```

An app that declares `x86_64-linux`, `arm64-darwin`, `aarch64-linux-gnu`, `arm64-linux`, and `x86_64-darwin-24` in its lockfile `PLATFORMS` needs the matrix above (x86_64-linux + arm64-darwin + aarch64-linux) to serve both its deploy target and its developers.

## Suggested steps

- [ ] Add `arm64-darwin` and `aarch64-linux` to the `cross-gem.yml` matrix
- [ ] Cut `v0.2.1` to exercise `release.yml` end to end
- [ ] Verify RubyGems lists the platform gems and that `bundle lock --add-platform x86_64-linux` resolves a native variant

cpb:
Closes #112.

## Purpose

- This PR adds two platforms to the precompiled gems.
- This PR corrects two defects that stopped those gems from working.
- This PR adds a test that installs each gem and calls it.
- This PR sets the version to 0.3.0.

## Platforms

- The cross-compile matrix now builds four platforms.
- The new platforms are `arm64-darwin` and `aarch64-linux`.
- The other two platforms are `x86_64-linux` and `x86_64-darwin`.
- Users on Apple Silicon and on arm64 Linux now get a precompiled gem.
- main already holds the `reference_time:` and `reference_zone:` work (#54, #102) that version 0.3.0 covers.

## Defect 1 — the aarch64-linux build did not link

- `Rake::ExtensionTask` always makes two builds.
- The first build is the cross build for `RUBY_TARGET`.
- The second build is a local build for the container's own Ruby.
- Each `rbsys/<platform>` image sets `RUST_TARGET` and `CARGO_BUILD_TARGET` for every process in the container.
- The local build gets these two variables too.
- On `aarch64-linux`, the local build makes aarch64 code.
- The local build then links that code with the host `gcc`.
- The code and the linker do not agree, and the link step fails.
- The local build writes the same file name as the cross build.
- The package step waits for that file, so the whole build stops.
- The correction is in the `Rakefile`.
- It sets the local build's Cargo target back to the true host triple.
- AGENTS.md holds the full record.

## Defect 2 — the gems raised TypeError on reference_time:

- To see this defect, download the gems from a **green** `cross-gem.yml` run.
- Install one gem, then call `Duckling.parse` with `reference_time:`.
- rb-sys reads Ruby's internal object layout at compile time.
- rb-sys reads that layout from the headers of the Ruby that builds the gem.
- rb-sys does not read that layout at run time.
- One binary therefore knows only the Ruby that built it.
- A 3.2 binary reads a 3.3 Time object incorrectly.
- magnus then refuses a true Time, and raises `TypeError: no implicit conversion of Time into Time`.
- `magnus-0.8.2/src/time.rs:238-241` holds that test.
- `require` still works, and parses that send no Ruby object into Rust still work.
- Only `reference_time:` and `reference_zone:` fail.
- The platform check, the architecture check and the `required_ruby_version` check all passed this gem.

Measured before the correction:

| Runtime | `require` | plain parse | `reference_time:` |
|---|---|---|---|
| ruby:3.2 linux/amd64 | ok | ok | **ok** |
| ruby:3.3 linux/amd64 | ok | ok | `TypeError` |
| ruby:3.4 linux/amd64 | ok | ok | `TypeError` |
| 3.3, 3.4 and 4.0 on macOS, `x86_64-darwin` gem | ok | ok | `TypeError` |

## Correction — one binary for each Ruby ABI

- Each precompiled gem now holds one binary for each Ruby ABI.
- `CROSS_RUBY_ABIS` in the `Rakefile` lists the ABIs: 3.2, 3.3, 3.4 and 4.0.
- The ABI list must agree in four places, and AGENTS.md names them.
- `lib/duckling.rb` loads the binary for the Ruby that runs.
- If that directory is not there, `lib/duckling.rb` loads the plain path.
- A source build and a local `rake compile` write the plain path.
- If both loads fail, `lib/duckling.rb` reports the first error.
- Alpine needs that first error: RubyGems gives an unversioned `-linux` gem to a musl runtime, so the binary is present but asks for glibc.
- `required_ruby_version` is now `>= 3.2.0, < 4.1.dev`.
- A Ruby above 4.0 gets the source gem, which builds its own binary.
- Without that limit, RubyGems installs a gem that holds no usable binary, and `require` fails.
- Each ABI adds about 3 MB to the packed gem.
- Remove the oldest ABI when that Ruby goes end of life, to keep the gem small.

## Ruby 4.0

- The gems hold a Ruby 4.0 binary.
- Without it, Ruby 4.0 users get the source gem.
- The source gem builds Rust code at install time.
- A standard Heroku build has no Rust tools, so that install fails.

## Two test suites for the gems

- The old checks only read the gem files, and never loaded one.
- `test/gem/packaged_gem_test.rb` reads the gems in `pkg/`.
- It checks the platform, `lib/duckling.rb`, the ABI set, `file(1)`'s architecture, and both limits of `required_ruby_version`.
- `test/gem/installed_gem_test.rb` installs a gem, loads it, and calls it.
- It makes sure the loaded binary comes from the gem's own ABI directory.
- It calls `reference_time:`, `reference_zone:` and one non-time dimension.
- The new `smoke` job runs that suite on a runner for each platform.
- The runners are `ubuntu-latest`, `ubuntu-24.04-arm`, `macos-15-intel` and `macos-latest`.
- The job runs one time for each ABI, so 16 jobs run.
- The job installs no Rust tools.
- If RubyGems refuses the precompiled gem and takes the source gem, the job fails.
- `release.yml` needs this whole workflow, so the smoke job controls the release.
- Neither suite takes an environment variable.
- Neither suite runs through rake or bundler, because both put this checkout's `lib/` ahead of the installed gem.
- Both run on the minitest CRuby bundles, because minitest 6 pulls in prism, a native extension a plain Ruby container cannot build.

## One place says what the gems are built for

- `cross_targets.rb` holds the ABI list, and each platform's architecture and runner.
- The `Rakefile` and both gem suites read it.
- `cross-gem.yml` cannot read it, because YAML cannot require Ruby.
- `cross-gem.yml` restates the matrix, and `test/duckling_ci_test.rb` fails when the copy drifts.
- That test used to read `CROSS_RUBY_ABIS` back out of the `Rakefile` with a regex over its source.

## The macos-13 runner

- GitHub retired the `macos-13` image in December 2025.
- A retired label does not fail the job.
- The job stays in the queue and never starts, so the run never reports.
- The smoke job now uses `macos-15-intel`.
- GitHub removes x86_64 macOS after `macos-15` retires, in fall 2027.

## Verification

- `bundle exec rake` is green: StandardRB, compile, and 153 tests.
- `cross-gem.yml` run 30953155493 is green: 20 of 20 jobs.
- Those 20 jobs are 4 cross-compiles and 16 smoke tests.
- A local `x86_64-linux` gem holds binaries for 3.2, 3.3, 3.4 and 4.0.
- Ruby 3.5 and Ruby 4.1 correctly get the source gem.
- The smoke suite passes on `ruby:3.2-slim`, `ruby:3.3-slim` and `ruby:3.4-slim`.
- Those containers hold no Rust tools, which is the Heroku condition.
- The smoke suite fails on the old gem, and reports the `TypeError`.
- That failure shows the test is not empty.

Verified again for the test-suite rework:

- `rake 'native_gem[aarch64-linux]'` from this worktree writes a complete gem.
- That gem holds `lib/duckling.rb` and binaries for all four ABIs.
- `ruby test/gem/packaged_gem_test.rb` passes its five aarch64 tests and skips the other fifteen.
- With `pkg/` empty, it fails rather than skipping everything.
- With a wrong architecture in `cross_targets.rb`, it fails and names the binary.
- Against the pre-fix single-ABI gem, it reports the missing 4.0 binary, the wrong ceiling, and the missing `lib/duckling.rb`.
- `ruby test/gem/installed_gem_test.rb` passes on that gem in `ruby:3.3-slim`, `--platform linux/arm64`, with no Rust tools.

## `rake 'native_gem[...]'` now works from a worktree

- `bin/worktree` sets up every branch here as a git worktree.
- A worktree's `.git` is a file that points outside the Docker mount.
- `git ls-files` failed there, so the gemspec file list came back empty.
- The gem then held only the binaries, and no `lib/*.rb`.
- The task now builds from a throwaway plain clone under `tmp/`.
- That clone's `.git` is a real directory inside the mount.
- The gem carries committed state, not the working tree.
- The clone does not reuse this checkout's Cargo target cache.

## Test plan

- [x] `bundle exec rake` is green.
- [x] `cross-gem.yml` is green for all four platforms and all four ABIs.
- [x] A precompiled gem works on every supported Ruby, on a Linux host with no Rust tools.
- [ ] Push the `v0.3.0` tag. **This step needs your agreement.**
- [ ] After the release, make sure `https://rubygems.org/api/v1/versions/duckling.json` lists the platform gems.
- [ ] After the release, make sure `bundle lock --add-platform aarch64-linux` finds a native gem.

## Release

- I did not push the `v0.3.0` tag.
- That tag publishes to RubyGems and makes a GitHub release.
- Push it only after this PR merges and you agree.

https://claude.ai/code/session_014mprvh4JwGXr3piDGGiWgA


.github/workflows/cross-gem.yml
```diff
@@ -32,15 +45,154 @@ jobs:
...
+                described = IO.popen(["file", "-b", binary], &:read).strip
+                unless described.include?(ENV.fetch("EXPECTED_ARCH"))
+                  failures << "#{File.basename(File.dirname(binary))}/#{File.basename(binary)} is #{described}, expected #{ENV["EXPECTED_ARCH"]}"
+                end
+                puts "#{File.basename(File.dirname(binary))}/#{File.basename(binary)}: #{described}"
+              end
+            end
+
+            abort("Gem verification failed:\n- #{failures.join("\n- ")}") if failures.any?
+            puts "#{spec.full_name} verified (#{spec.required_ruby_version})"
```
cpb: **issue:** extract this to a script

.github/workflows/cross-gem.yml
```diff
@@ -32,15 +45,154 @@ jobs:
...
+        include:
+          - platform: x86_64-linux
+            runner: ubuntu-latest
+          - platform: aarch64-linux
+            runner: ubuntu-24.04-arm
+          # Intel macOS. GitHub retired macos-13 in December 2025, and a
+          # retired label does not fail — the job queues forever. GitHub drops
+          # x86_64 macOS entirely once macos-15 retires, currently announced
+          # for fall 2027; move to macos-26-intel before then, and reconsider
+          # shipping an x86_64-darwin gem at all once no runner can test it.
```
cpb: **issue:** this is chain of thought, not useful for a operator in the future

cpb (commented):
Review all code comment and agent contributions for chain of thought rather than documenting the current state for a future operator.

cpb:
Addressed in 54a9f88 — audited every comment and AGENTS.md passage this PR adds, against the "Code comment conventions" rule that comments are long-lived documentation rather than a transcript of the session that wrote them.

Rewritten from chain-of-thought to current state:

- AGENTS.md: **"One binary per platform is not enough — this was shipped wrong once"** → **"Each Ruby ABI needs its own binary"**, then the mechanism and the detection rule.
- AGENTS.md: `smoke` "is the only place the built gems are ever loaded and called — static inspection alone *once let* a gem that raised `TypeError` pass every check" → an ABI mismatch passes every metadata check and fails on the first call, so nothing earlier in the pipeline can catch it.
- AGENTS.md, thin-gem bullet: "*only inbound `Time` conversion actually broke* … Don't" → states the property (an ABI mismatch does not break every call) and the rule (keep one binary per ABI), without recounting what was investigated.
- `cross-gem.yml`: "State the two ways it goes wrong, *rather than pinning the string the Rakefile happens to build*" → describes what the checks test and why a fixed string is the wrong assertion.
- `cross-gem.yml`: the Intel runner comment led with "GitHub retired macos-13 in December 2025" → leads with the requirement (this label must name a live runner image) and the failure mode (a retired label queues forever rather than failing).
- `lib/duckling.rb`: "One binary cannot serve every Ruby … A binary built against 3.2 misreads a 3.3 object" → "Load the one for the running Ruby", then the mechanism in general terms.
- `Rakefile`: "One binary per platform is not enough" → "A binary understands only the Ruby that built it"; "a host triple *we* cannot read" → "a host triple that cannot be read".

Two corrections found while auditing:

- The AGENTS.md `cross-gem.yml` entry still listed `macos-13`, which no longer exists. It now names `macos-15-intel`.
- Both the `Rakefile` and `cross-gem.yml` comments said the ABI list must agree with one other place. It is four (`CROSS_RUBY_ABIS`, `ruby-versions`, `EXPECTED_ABIS`, and the smoke matrix). Each comment now names the other three.

`bundle exec rake` is green (150 runs, 0 failures) and `cross-gem.yml` parses.

.github/workflows/cross-gem.yml
cpb: Extracted to `.github/scripts/verify-gem.rb` in 9a61a3d. The step is now:

```yaml
      - name: Verify gem metadata and architecture
        env:
          BUILT_GEM: ${{ steps.cross-gem.outputs.gem-path }}
          EXPECTED_PLATFORM: ${{ matrix.platform }}
          EXPECTED_ARCH: ${{ matrix.expected_arch }}
          EXPECTED_ABIS: "3.2,3.3,3.4,4.0"
        run: ruby .github/scripts/verify-gem.rb
```

It takes the same environment variables, so it runs by hand against any built gem — the header carries the one-liner. That also puts it beside `smoke-gem.rb`, and the two now split cleanly: `verify-gem.rb` examines the gem file, `smoke-gem.rb` installs the gem and calls it.

Exercised both ways locally before pushing:

- Against a 4-ABI gem whose `required_ruby_version` was capped at `< 3.5.dev`, it reports `required_ruby_version >= 3.2.0, < 3.5.dev excludes Ruby 4.0, which this gem carries a binary for` and exits 1 — real drift between the built ABIs and the metadata.
- Against the old single-ABI gem, it reports `admits Ruby 4.1.0, which this gem carries no binary for` and `carries binaries for [], expected ["3.2", "3.3", "3.4", "4.0"]`, and exits 1.

`standardrb` is clean on it, and it now lints on every change like any other Ruby file in the repo, which it could not do inside the YAML string.

.github/workflows/cross-gem.yml
cpb: Rewritten in 54a9f88. It led with the history of the label rather than what a future operator needs. Now:

```yaml
          # Intel macOS. This label must name a live runner image: GitHub does
          # not fail a job that asks for a retired one, it leaves the job
          # queued forever, so the run never reports at all.
          #
          # GitHub drops x86_64 macOS entirely once macos-15 retires,
          # announced for fall 2027. Move to macos-26-intel before then. Once
          # no runner can execute an x86_64-darwin gem, reconsider shipping
          # one.
```

The requirement comes first, then the failure mode that makes it hard to diagnose, then the next action and its deadline.

I applied the same read to every comment and AGENTS.md passage this PR adds — six others were narrating the session rather than documenting the system. Full list in [this comment](https://github.com/cpb/duckling/pull/113#issuecomment-5185167607). Two stale facts fell out of that audit: AGENTS.md still named `macos-13`, and both the `Rakefile` and workflow comments said the ABI list must agree with one other place when it is four.

PR #118, opened 2026-08-08
Drop the tzinfo-data runtime dependency, and prove the fallback is covered

cpb:
Closes #117.

`tzinfo` already prefers the `tzinfo-data` gem when it is installed and falls back to the host's zoneinfo files otherwise, so depending on it forced bundled tz data on every consumer to serve the ones who want it. Removing the dependency is one line.

The rest of this PR exists because **the suite could not tell which tz database it ran against** — it passed identically on both — and dropping the dependency is precisely a change in which database we run against.

Shipping as **v0.4.0** rather than 0.3.1: the gem's documented API is unchanged, but which zone identifiers resolve on a stock host is not.

## What consumers see

On a stock Debian/Ubuntu host, roughly a hundred backward-compat identifiers (`US/Eastern`, `Europe/Kiev`, `Japan`, …) stop resolving. Canonical names are unaffected. Both remedies restore the previous behavior exactly — `gem "tzinfo-data"` or `apt install tzdata-legacy` — and the error names the database that answered, how many identifiers it has, and the remedy, so a valid IANA name doesn't read as a typo.

On a host with *no* tz database at all (scratch, distroless), `reference_zone:` raises `Duckling::TZDataUnavailable` naming both fixes, rather than a raw tzinfo error mentioning neither this gem nor the keyword. Every other keyword works without a tz database.

One more, found while building the Alpine environment: a **musl** consumer building the source gem needs `libclang` + clang's resource headers for rb-sys's bindgen, and on Alpine also `BINDGEN_EXTRA_CLANG_ARGS=-I<resource-dir>` (Alpine splits the headers away from the library's search path). The `tz-containers` job pins the working recipe; no precompiled gem targets musl (`cross_targets.rb`), so the source build is the path musl consumers actually take.

## How the suite tells databases apart

The two databases disagree on independent axes — **modelling** (rearguard tzdata strips negative DST; backward-compat links live in `tzdata-legacy`) and **vintage** (a pinned gem or unpatched host can predate a rule change and answer with old rules rather than raising). Four mechanisms, none of them a skip manifest:

1. **Environments.** `DUCKLING_TZINFO_DATA` (Gemfile) and `DUCKLING_ZONEINFO_DIR` (`test_helper.rb`) select the database; seven environments run in CI (see below). An environment is a *checked state*, not just a matrix slot.
[...]

Issue #29, opened 2026-07-01
Expand CI test matrix: Claude Code Web baseline + Ruby 3.4/latest + latest Rust


cpb:
## Goal

Expand CI's test matrix beyond the single `Ruby 3.3.6` / pinned-Rust combination to also cover Ruby `3.4`, the latest available Ruby, and the latest available Rust — verifying the gem against both the Claude Code Web baseline and newer toolchains.

## Background

Main's CI currently runs a single combination: Ruby `3.3.6` with a Rust toolchain pinned/tracked to what's available in Claude Code Web (see #28). That's the right baseline for "what an agent working in Claude Code Web can run without installing anything extra," but it doesn't catch forward-compatibility breaks — e.g. a native extension API that stops working on a newer Ruby ABI, or Rust code that only compiles on an older `rustc` and breaks on the newest stable release.

Raised during PR #3 review (research/planning PR for issue #1). Not blocking PR #3 (docs-only).

## Acceptance criteria

- [ ] CI matrix includes at least: the Claude Code Web baseline (Ruby 3.3.6 + pinned Rust, per #28), Ruby 3.4, the latest released Ruby, and the latest stable Rust
- [ ] Matrix combinations that are expected to fail or are not yet supported are either excluded with a documented reason, or allowed to fail without blocking the required baseline job
- [ ] CI clearly reports which matrix entry is the "baseline" (must pass to merge) vs. "informational" (forward-compat signal), if not all combinations are required

## Out of scope

- Guaranteeing support for every historical Ruby/Rust version — this is about forward compatibility, not backward compatibility
- Automated tracking of "latest Ruby"/"latest Rust" (a periodic manual bump, or a scheduled job that opens a PR when a new version is released, are both acceptable approaches — implementation detail for whoever picks this up)

## Notes

Depends on / builds on #28 (pinning the Claude Code Web baseline Rust version) and complements #26/#27 (local + Claude Code Web setup scripts).

PR #120, opened 2026-08-09
Run the suite in the tz-divergent container environments in CI

cpb:
Extracts the CI environment coverage from #118, narrowed to what v0.3.0 can honestly cover: **no environment contracts, and no database selection** — no `DUCKLING_ZONEINFO_DIR` override, no `DUCKLING_TZINFO_DATA`. The suite runs exactly as the gem ships.

## What runs

A new `tz-containers` job with the two host environments no GitHub runner image provides:

| environment | image | what the host carries |
|---|---|---|
| `source build: Debian trixie + tzdata-legacy` | `ruby:3.4-trixie` | vanguard tzdata (negative-DST modelling intact) + the backward-compat links Debian split into `tzdata-legacy` |
| `source build: Alpine/musl` | `ruby:3.4-alpine` (digest-pinned) | vanguard tzdata on musl; also the suite's only musl source-build coverage, since no precompiled gem targets musl |

The entries are named for their present-tense coverage — on the tz axis these legs prove a negative today, so a red leg means a toolchain/build break, not tz handling. #118 renames them when they become genuine tz environments.

## Gating status (deliberate)

Not merge-blocking (branch protection requires only the `baseline` context) but **release-blocking**: `release.yml` runs this workflow as its `ci` reusable-workflow job, and `cross_gems`, `benchmark`, and `publish` all `needs: ci`, and a reusable-workflow job succeeds only if every job in the called workflow does. That is the intended gate — don't cut a release the musl source build can't serve — but it is a behavioral change to releases, and it makes these the only release-gating legs that hit external package repositories at release time (every other job uses preinstalled runner software or pinned actions), so the install steps take precautions: the Alpine image is digest-pinned (the floating tag silently rebases across Alpine releases, which drop versioned `clang*` packages as they roll) and `bundle install` retries.

## Why this is the coverage v0.3.0 can have

On the tz axis these jobs prove a negative, and that is the point: both hosts carry tz data materially different from the ubuntu runner's, and the suite is green regardless — because the gem depends on `tzinfo-data` and tzinfo prefers it, so the host's database never reaches the suite. **Providing `tzinfo-data` as a dependency is what resolves these environment concerns today**, and these jobs pin that down (alongside the real, present-tense musl/trixie build coverage).

What deliberately isn't here:

- **No overriding the database.** With `tzinfo-data` a gemspec dependency, pointing the suite at the host's zoneinfo would require forcing the datasource; that machinery belongs to the change that actually makes the host database reachable.
- **No `DUCKLING_TZINFO_DATA` / Gemfile / gemspec / lockfile changes.** The bundle is the default bundle in every leg.
- **No environment contracts, capability gates, build scripts, or fixture zones** — a stale or linkless zoneinfo tree that nothing reads would be theater, since the suite passes identically either way while the gem answers.
- The pinned `tzinfo-data-1.2022.7` leg can't exist here regardless: the pin conflicts with the gemspec's `~> 1.2024`.

## Then #118 shows the issues

#118 rebases on top of this and drops the dependency. Its diff then carries everything that reveals and supports what the environments differ in: the datasource selection (`DUCKLING_TZINFO_DATA`, `DUCKLING_ZONEINFO_DIR`), the full set of tz environments (baseline system-zoneinfo + linkless steps, the two stale legs in `timezones`, these containers pointed at their own tz data), the environment contracts (`test/environments/`) that assert each state positively, the capability-gated tests, the `zic` fixture zones, the unknown-identifier diagnosis, and the v0.4.0 bump.

## Verification

Both legs run locally in Docker for this revision: 153 runs, 0 failures each — the stock v0.3.0 suite, green on vanguard+links and vanguard+musl hosts alike. The Alpine run also proves the musl source build with the documented rb-sys bindgen recipe (`clang22-libclang` + `clang22-headers` + `BINDGEN_EXTRA_CLANG_ARGS`).


AGENTS.md
```diff
@@ -41,7 +41,7 @@ If you notice this file describing a not-yet-built piece as current, or vice ver
...
...`name: "Ruby 3.3.6"`, hardcoded Ruby 3.3.6 + a Rust toolchain pinned to track the Claude Code Web sandbox — the only entry branch protection on `main` requires, matched by that exact `name` string), `tz-containers` (`needs: baseline`, a 2-entry matrix of containerized host environments no runner image provides — Debian trixie + `tzdata-legacy` and Alpine; both carry tz data materially different from the runner's, and the suite is green regardless because the bundled `tzinfo-data` is what tzinfo reads — the jobs' present-tense coverage is the musl/trixie source builds, and they become genuine tz environments when #118 drops that dependency) and `informational` (`needs: baseline`, a 3-entry matrix of forward-compat signals allowed to fail — Ruby 3.4, Ruby latest, Rust latest). Because `informational` declares `needs: baseline`, GitHub Actions automatically skips it when `baseline` fails, instead of always burning CI minutes running all 4 entries in parallel regardless of the baseline's outcome (see #98). Rename the `baseline` job's `name` without updating the required check in the same PR and merges silently have no required check at all. Runs for every push to `main` and every PR. |
```
cpb: This rewritten line is now internally stale in two ways:

1. **"all 4 entries"** — with the 2-entry `tz-containers` matrix, the fan-out after `baseline` is now 6 entries (1 baseline + 2 tz-containers + 3 informational), not 4.
2. **"Because `informational` declares `needs: baseline`…"** — this now attributes the skip-on-baseline-failure behavior to only one of the two jobs that have it. `tz-containers` also declares `needs: baseline` and gets skipped the same way.

Given this file's own freshness rule at the top ("If you notice this file describing a not-yet-built piece as current, or vice versa…"), the one line this PR touches shouldn't merge already out of date. Suggest "all 6 entries" and naming both jobs (or rewording to "the downstream jobs").

.github/workflows/main.yml
```diff
@@ -49,6 +49,80 @@ jobs:
...
+    runs-on: ubuntu-latest
+    name: ${{ matrix.name }}
+    container: ${{ matrix.image }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: "tz: system zoneinfo + tzdata-legacy (Debian)"
+            image: ruby:3.4-trixie
+            setup: apt-get update -qq && apt-get install -y --no-install-recommends build-essential cargo rustc git tzdata tzdata-legacy libclang-dev
```
cpb, line 115: Verified empirically on today's `ruby:3.4-trixie` (Debian 13.6): this install line works, and the full leg passes — `bundle install` + `bundle exec rake test` → 153 runs, 0 failures.

But note what trixie ships: **rustc 1.85.0, frozen for the life of Debian 13**. `ext/duckling/Cargo.toml` is `edition = "2024"` with no `rust-version` field, so the crate's floor is *exactly* 1.85 — this leg is already running on the minimum compiler it will ever have. Unlike `baseline`'s pinned 1.94.1, there is no upgrade path here: the first crate change requiring >1.85 (let-chains need 1.88, etc.) turns this leg red with a confusing parse error while every other job stays green, and nothing records that constraint.

Cheap fixes, either or both: add `rust-version = "1.85"` to `Cargo.toml` (converts the future failure into a clean MSRV error from cargo), or a comment here stating that trixie's frozen rustc is the de facto MSRV floor for this leg.

.github/workflows/main.yml
```diff
@@ -49,6 +49,80 @@ jobs:
...
+          # libclang; and Alpine splits clang's resource headers
+          # (stdckdint.h, which ruby-3.4's headers include) away from the
+          # library's default search path, so bindgen is pointed at them
+          # explicitly. Any musl consumer building the source gem needs the
+          # same three things. #118 adds tzdata-utils (zic) for its
+          # fixture-zone tests.
+          - name: "tz: Alpine vanguard zoneinfo (musl source build)"
+            image: ruby:3.4-alpine
+            setup: |
+              apk add --no-cache build-base cargo rust git tzdata clang22-libclang clang22-headers
```
cpb, line 136: Verified on today's `ruby:3.4-alpine` (Alpine 3.24.1): the `clang22-*` packages exist, BusyBox `find` supports `-print -quit`, and the guard + `BINDGEN_EXTRA_CLANG_ARGS` export work end-to-end (full leg green: 153 runs, 0 failures).

The fragility is the floating tag: `ruby:3.4-alpine` will silently rebase onto newer Alpine releases, and Alpine drops older versioned clang packages as it rolls. When the tag lands on an Alpine without `clang22-libclang`/`clang22-headers`, this leg fails at `apk add` with zero changes in this repo. The `find /usr/lib/llvm*` is version-agnostic — only the package names aren't.

Options: pin the image by digest (consistent with the SHA-pinned actions above), or resolve the names dynamically, e.g. `apk search -q 'clang*-libclang' | sort -V | tail -1`. (I checked — there is no unversioned `clang-libclang` package to switch to, so the versioned names are currently the only option; that makes digest-pinning the more honest fix.)

.github/workflows/main.yml
```diff
@@ -49,6 +49,80 @@ jobs:
...
+  # tzinfo-data and tzinfo prefers it — the host's database never reaches the
+  # suite. #118 drops that dependency, points the suite at the host's
+  # database in jobs like these, and adds the environment contracts that pin
+  # what each state then means.
+  #
+  # Container jobs, so the image's own Ruby runs and ruby/setup-ruby does not
+  # apply (it has no musl support anyway); the toolchain comes from the
+  # image's package manager, git included — which is why the install step
+  # precedes checkout.
+  tz-containers:
```
cpb, line 97: This job is not merge-blocking (branch protection requires only the `baseline` context) — but as of this PR it **is release-blocking**, and the PR body doesn't mention it. `release.yml` runs this workflow as its `ci` reusable-workflow job, and `cross_gems`, `benchmark`, and `publish` all `needs: ci`. A reusable-workflow job succeeds only if every job in the called workflow does, so an apt/alpine mirror flake or a musl regression now blocks publishing a tag.

That makes these the only release-gating legs that hit external package repositories at release time — every other job in the pipeline uses preinstalled runner software or pinned actions. Maybe that's the intended gate (arguably correct — you don't want to cut a release the musl build can't serve), but it's a behavioral change to releases smuggled in with a CI-coverage PR, and unlike `informational` (whose comment states "not required to merge (see issue #29)") nothing here records the intended gating status. Suggest stating it explicitly in both the comment block and the PR body.

.github/workflows/main.yml
```diff
@@ -49,6 +49,80 @@ jobs:
...
+    needs: baseline
+    runs-on: ubuntu-latest
+    name: ${{ matrix.name }}
```
cpb, line 100: The `tz: …` names describe an environment the suite never reads. Per this PR's own (commendably honest) framing, `tzinfo-data` wins and the host database never reaches the suite — on the tz axis these legs prove a negative. So if `tz: Alpine vanguard zoneinfo (musl source build)` goes red tomorrow, the cause will be a toolchain/build break, not tz handling, and the check name will point whoever's debugging at the wrong layer.

Consider naming for the present-tense coverage (e.g. `source build: Alpine/musl`, `source build: Debian trixie + tzdata-legacy`) and letting #118 rename them when they become genuine tz environments. That feels more consistent with the PR's own standard — it declined to ship fixture zones as "theater," but the names assert tz coverage that doesn't exist yet.

.github/workflows/main.yml
```diff
@@ -49,6 +49,80 @@ jobs:
...
+              echo "BINDGEN_EXTRA_CLANG_ARGS=-I$(dirname "$headers")" >> "$GITHUB_ENV"
+
+    steps:
+      - name: Install system packages
+        run: ${{ matrix.setup }}
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Install dependencies
+        run: bundle install --jobs 4
```
cpb: Nit: `bundle install --jobs 4` with no `--retry` and no cache. Every run re-downloads the full gem set and cold-compiles the native extension; a transient rubygems blip fails the leg (and, per the release-gating comment above, potentially a release). `baseline` gets retries and caching for free via `ruby/setup-ruby`'s `bundler-cache`; these legs get neither. `--retry 3` is cheap insurance even if caching stays out of scope.

cpb (commented):
Adversarial review. Before commenting I re-ran both legs end-to-end in Docker at this revision — Debian trixie leg: 153 runs, 0 failures; Alpine/musl leg: 153 runs, 0 failures — so the PR's verification claim holds, and the package names / BusyBox `find -quit` / clang-header guard all behave as the comments describe.

Things I checked that turned out **fine** (no comment left): `actions/checkout` on musl — the runner ships `node20_alpine` (from `actions/alpine_nodejs`), so JS actions work in the Alpine container (a glibc node indeed fails there; I verified the failure mode and then found the runner's musl build). The `find` guard's reasoning is correct (`find` exits 0 on no match; `dirname ""` is `.`). `Gemfile.lock` lacking a musl platform is harmless here — bundler re-resolves in non-frozen mode. `cross_targets.rb` confirms no precompiled gem targets musl.

The inline findings, in rough severity order: the AGENTS.md line this PR rewrites is already stale ("all 4 entries"); the trixie leg is pinned to rustc 1.85 forever while the crate's floor is exactly 1.85; the hardcoded `clang22-*` package names will break when the floating `ruby:3.4-alpine` tag rebases; this job silently becomes **release-blocking** via `release.yml`'s `needs: ci` without that being stated anywhere; the `tz:` check names assert coverage the suite doesn't exercise yet; and a `--retry` nit on `bundle install`.

AGENTS.md
cpb: Fixed in two passes: 7b885f0 named both jobs and corrected the count, and ce75ba7 reworded it to count just the downstream matrix entries — "Because `tz-containers` and `informational` both declare `needs: baseline`, GitHub Actions automatically skips all 5 of their matrix entries…".

.github/workflows/main.yml
cpb: Both, in 7b885f0: `rust-version = "1.85"` is now in `ext/duckling/Cargo.toml` (edition 2024 already floored the compiler there; naming it converts the future failure into a clean MSRV error from cargo), and the Debian matrix entry carries a comment stating that trixie's frozen rustc 1.85.0 is the de facto floor for the leg with no upgrade path. Verified the crate still compiles with rustc 1.85.0 exactly.

.github/workflows/main.yml
cpb: Stated explicitly in all three places in 7b885f0: the tz-containers comment block ("not merge-blocking… but release-blocking… That is the intended gate: don't cut a release the musl source build can't serve. It does make these the only release-gating legs that hit external package repositories at release time"), the AGENTS.md row, and the PR body (new "Gating status (deliberate)" section).

.github/workflows/main.yml
cpb: Digest-pinned in 7b885f0: `ruby:3.4-alpine@sha256:c5a5064d…` (3.4.10-alpine3.24), with a comment explaining why (the floating tag silently rebases across Alpine releases, which drop the versioned clang packages) and how to bump (`docker buildx imagetools inspect ruby:3.4-alpine`, then re-verify the clang package names). The trixie tag names a frozen Debian release, so it doesn't have the rebase failure mode. Re-ran the leg end-to-end against the pinned digest: green.

.github/workflows/main.yml
cpb: Renamed in 7b885f0: `source build: Debian trixie + tzdata-legacy` and `source build: Alpine/musl` — named for the present-tense coverage, so a red leg points at the toolchain/build layer, which is what can actually break here today. The comment block says #118 renames them when they become genuine tz environments.

.github/workflows/main.yml
cpb: Added `--retry 3` in 7b885f0, with a comment noting these legs get neither bundler-cache nor retries otherwise, and that a transient rubygems blip would otherwise fail a release-gating leg. Caching stays out of scope, as you framed it.

PR #255, opened 2026-08-20
Add the wafer-inc/duckling EN corpus as end-to-end parse coverage

### LINKED ISSUE #157: Add wafer-inc/duckling EN corpus as end-to-end parse coverage (1,656 cases)
## Goal

Add the wafer-inc/duckling per-dimension EN test corpora — **1,656 `check_*` assertions across all 13 dimensions** — as automated end-to-end coverage of `Duckling.parse`'s tagged-Hash output, via a checked-in extracted fixture plus a minitest runner.

## Background

Our time tests already mirror the wafer-inc/pyduckling EN corpus *by hand* (see the comment in `test/duckling_parse_time_basic_test.rb`) — partially and manually. The upstream repo's `tests/*_corpus.rs` files are the full ported Haskell corpus and are far more complete.

Spike result: I extracted every `check_*` call site from the upstream corpus and replayed all of them through `Duckling.parse` on main @ `a6ee4f6`:

| Corpus | Cases | Result |
|---|---|---|
| time (naive/instant/interval/open-interval/with-context) | 1,034 | ✅ all pass |
| amount_of_money | 141 | ✅ all pass |
| numeral | 105 | ✅ all pass |
| duration | 83 | ✅ all pass |
| volume | 54 | ✅ all pass |
| distance | 51 | ✅ all pass |
| quantity | 50 | ✅ all pass |
| temperature | 33 | ✅ all pass |
| ordinal | 32 | ✅ all pass |
| credit_card_number | 22 | ✅ all pass |
| phone_number | 19 | ✅ all pass |
| url | 19 | ✅ all pass |
| email | 13 | ✅ all pass |
| **total** | **1,656** | **1,656 / 1,656** |

This corpus pins exactly the behaviors the unified `serde_magnus` shape (#90/#93/#97) and the #100 Data-object proposal rely on: per-dim tagged hashes, grain symbols, real `Time` leaves, open-interval `nil` bounds (13 explicit `check_time_open_interval_after/before` cases), `Measurement` `Value`/`Interval` discrimination, and quantity products — plus two non-default reference-time contexts (Saturday / Friday-evening) for weekend resolution.

It also pins **deadline synthesis**: the crate's `by <time>` rule builds a *closed* interval `[now, named]` (`TimeForm::Interval(Now, t)`, `src/dimensions/time/en.rs`), and all 13 `"by …"` corpus cases expect `from` == the reference instant exactly. So "both bounds present" ≠ "both bounds stated in the text" — a distinction consumer code has to detect by comparing `from` against the reference time.

**Matcher-semantics trap (measured):** the Rust measurement checkers accept an `Interval` when *either* bound matches value+unit ("matching original test behavior"). A naive `Value`-only matcher reports 93 false failures on interval phrasings ("between 10 and 20 dollars"). The runner below mirrors the upstream semantics exactly; the real implementation must too.

**License:** wafer-inc/duckling is BSD-3-Clause, and its corpora are ported from facebook/duckling's Haskell corpora (also BSD). Vendoring derived fixture data is fine with the notice preserved — same as we already do for the bundled crate.

## Proposal

- `script/extract_wafer_corpus.rb` — the extractor below, parameterized on repo paths (spike hardcodes `/tmp`), producing `test/fixtures/wafer_corpus.json` (checked in; header records the upstream sha it was extracted from).
- `test/duckling_wafer_corpus_test.rb` — the runner below, minitest-ified (one test per case, or per file with per-case subassertions).
- A `rake corpus:refresh` task (or `bin/` script) that clones wafer-inc/duckling at a recorded ref and regenerates the fixture.
- **Supplementary cases for a gap found during the spike:** the `by <time>` deadline mechanism is pinned upstream, but no `by <named day/date>` case exists — e.g. `"by Friday"`, `"by Friday July 3, 2026"`, where the `to` bound is a `Naive` calendar day at `:second` grain (verified live on main). Add a small number of these as clearly-marked local additions to the fixture.

## Acceptance criteria

- [ ] Fixture + runner in tree; the 1,656 cases run under the normal test task
- [ ] Matcher semantics mirror upstream exactly: `.any` over entities, either-bound interval acceptance for measurements, 0.01 float tolerance, wall-clock component comparison, corpus reference times (2013-02-12 04:30 −02:00 default; Saturday 2013-02-09 10:00 and Friday-evening 2013-02-08 20:00 contexts)
- [ ] Fixture records the upstream sha; refresh task documented
- [ ] BSD-3-Clause notice for the vendored derived data
- [ ] Runs in CI (~1,700 short-string parses — seconds)

## Out of scope

- `pending_corpus.rs` (known-pending upstream; optionally imported later as skip-marked tests to track upstream progress)
- `extreme_inputs_regression.rs` / `time_chaos_regression.rs` (crash/sanity regression — `assert_time_point_sane`-style, no value expectations; a separate concern)
- `tests/haskell_upstream/` (reference data, not tests)
- Non-EN locales (upstream corpus coverage is EN-only)

## Scripts (spike quality, verbatim from the verification run)

Reproduce: `git clone --depth 1 https://github.com/wafer-inc/duckling /tmp/wafer-duckling && ruby extract.rb && bundle exec ruby -rduckling run.rb`

<details><summary><code>extract_wafer_corpus.rb</code> — corpus .rs files → JSON manifest</summary>

```ruby
# Extracts check_* call sites from wafer-inc/duckling corpus files into JSON.
require "json"

CHECKS = %w[
  time_interval_with_context time_open_interval_after time_open_interval_before
  time_naive time_instant time_interval no_time
  quantity_with_product quantity
  money cc no_cc distance duration no_duration email no_email numeral ordinal
  phone no_phone temperature url no_url volume
].join("|")

def parse_arg(s)
  case s
  when /\A"(.*)"\z/m
    $1.gsub(/\\u\{([0-9a-fA-F]+)\}/) { [$1.to_i(16)].pack("U") }
      .gsub(/\\(.)/) { { "n" => "\n", "t" => "\t", "r" => "\r", '"' => '"', "\\" => "\\", "0" => "\0" }.fetch($1, $1) }
  when /\Adt\((.*)\)\z/m
    ["dt", *$1.split(",").map { |x| x.strip.to_i }]
  when /\Alocal_datetime\((.*)\)\z/m
    ["ldt", *$1.split(",").map { |x| x.strip.to_i }]
  when /\A-?\d+\z/ then s.to_i
  when /\A-?\d+\.\d+\z/ then s.to_f
  else ["raw", s]
  end
end

cases = []
Dir["/tmp/wafer-duckling/tests/*_corpus.rs"].each do |file|
  src = File.read(file)
  pos = 0
  while (m = /(?<!fn )\bcheck_(#{CHECKS})\s*\(/.match(src, pos))
    name = m[1]
    # skip matches on commented-out lines
    line_start = src.rindex("\n", m.begin(0)) || 0
    if src[line_start...m.begin(0)].include?("//")
      pos = m.end(0)
      next
    end
    # consume balanced parens, string-literal aware
    i = m.end(0) - 1
    depth = 0; j = i; in_str = false; esc = false
    while j < src.length
      c = src[j]
      if in_str
        if esc then esc = false
        elsif c == "\\" then esc = true
        elsif c == '"' then in_str = false end
      else
        case c
        when '"' then in_str = true
        when "(" then depth += 1
        when ")" then depth -= 1; break if depth.zero?
        end
      end
      j += 1
    end
    argstr = src[(i + 1)...j]
    pos = j + 1

    # split top-level commas
    args = []; depth = 0; in_str = false; esc = false; cur = +""
    argstr.each_char do |c|
      if in_str
        cur << c
        if esc then esc = false
        elsif c == "\\" then esc = true
        elsif c == '"' then in_str = false end
      else
        case c
        when '"' then in_str = true; cur << c
        when "(" then depth += 1; cur << c
        when ")" then depth -= 1; cur << c
        when ","
          if depth.zero? then args << cur.strip; cur = +""
          else cur << c end
        else cur << c
        end
      end
    end
    args << cur.strip unless cur.strip.empty?

    parsed = args.map { |a| parse_arg(a) }
    if name == "time_interval_with_context"
      ctx = src[0...m.begin(0)].scan(/let ctx = make_(\w+)_context/).flatten.last
      parsed[1] = ["ctx", ctx]
    end
    cases << { file: File.basename(file), check: name, text: parsed[0], args: parsed[1..] }
  end
end

File.write("/tmp/corpus_cases.json", JSON.pretty_generate(cases))
puts "#{cases.size} cases"
puts cases.group_by { |c| c[:file] }.transform_values(&:size).sort.to_h
```

</details>

<details><summary><code>run_wafer_corpus.rb</code> — replay manifest against <code>Duckling.parse</code></summary>

```ruby
# Runs the extracted wafer-inc/duckling corpus cases against this gem's
# Duckling.parse and checks the same expectations the Rust corpus checks.
require "json"

CASES = JSON.parse(File.read("/tmp/corpus_cases.json"))

REF = Time.new(2013, 2, 12, 4, 30, 0, "-02:00")
CTX = {
  "saturday" => Time.new(2013, 2, 9, 10, 0, 0, "-02:00"),
  "friday_evening" => Time.new(2013, 2, 8, 20, 0, 0, "-02:00"),
}

DIM = {
  "money" => "amount-of-money", "cc" => "credit-card-number", "distance" => "distance",
  "duration" => "duration", "email" => "email", "numeral" => "number", "ordinal" => "ordinal",
  "phone" => "phone-number", "quantity" => "quantity", "quantity_with_product" => "quantity",
  "temperature" => "temperature", "url" => "url", "volume" => "volume",
  "time_naive" => "time", "time_instant" => "time", "time_interval" => "time",
  "time_open_interval_after" => "time", "time_open_interval_before" => "time",
  "time_interval_with_context" => "time",
  "no_cc" => "credit-card-number", "no_duration" => "duration", "no_email" => "email",
  "no_phone" => "phone-number", "no_time" => "time", "no_url" => "url",
}.freeze

def wc(t) = [t.year, t.month, t.day, t.hour, t.min, t.sec]
def dt_parts(a) = a[1..]
def leaf(pt) = pt && (pt[:Naive] || pt[:Instant])

# Mirrors the Rust corpus checkers: accept the Value variant, or an Interval
# where EITHER bound matches value+unit ("matching original test behavior").
def measurement_match(m, val, unit)
  return false unless m
  if (s = m[:Value])
    (s[:value] - val).abs < 0.01 && s[:unit] == unit
  elsif (i = m[:Interval])
    [i[:from], i[:to]].compact.any? { |b| (b[:value] - val).abs < 0.01 && b[:unit] == unit }
  else
    false
  end
end

def measurement(v, tag, val, unit)
  measurement_match(v[tag], val, unit)
end

def match?(check, args, entities)
  vals = entities.map { |e| e[:value] }
  case check
  when "numeral" then vals.any? { |v| n = v[:Numeral]; n && (n - args[0]).abs < 0.01 }
  when "ordinal" then vals.any? { |v| v[:Ordinal] == args[0] }
  when "email" then vals.any? { |v| v[:Email] == args[0] }
  when "url" then vals.any? { |v| u = v[:Url]; u && u[:value] == args[0] && u[:domain] == args[1] }
  when "phone" then vals.any? { |v| v[:PhoneNumber] == args[0] }
  when "cc" then vals.any? { |v| c = v[:CreditCardNumber]; c && c[:value] == args[0] && c[:issuer] == args[1] }
  when "money" then vals.any? { |v| measurement(v, :AmountOfMoney, args[0], args[1]) }
  when "distance" then vals.any? { |v| measurement(v, :Distance, args[0], args[1]) }
  when "volume" then vals.any? { |v| measurement(v, :Volume, args[0], args[1]) }
  when "temperature" then vals.any? { |v| measurement(v, :Temperature, args[0], args[1]) }
  when "quantity", "quantity_with_product"
    vals.any? do |v|
      q = v[:Quantity]
      next false unless q
      next false if check == "quantity_with_product" && q[:product] != args[2]
      measurement_match(q[:measurement], args[0], args[1])
    end
  when "duration"
    vals.any? { |v| d = v[:Duration]; d && d[:value] == args[0] && d[:grain] == args[1].to_sym }
  when "time_naive"
    vals.any? do |v|
      p = v.dig(:Time, :Single, :value, :Naive)
      p && wc(p[:value]) == dt_parts(args[0]) && p[:grain] == args[1].to_sym
    end
  when "time_instant"
    vals.any? do |v|
      p = v.dig(:Time, :Single, :value, :Instant)
      p && wc(p[:value]) == dt_parts(args[0]) && p[:grain] == args[1].to_sym
    end
  when "time_interval", "time_interval_with_context"
    vals.any? do |v|
      i = v.dig(:Time, :Interval)
      next false unless i && i[:from] && i[:to]
      f = leaf(i[:from]); t = leaf(i[:to])
      wc(f[:value]) == dt_parts(args[0]) && wc(t[:value]) == dt_parts(args[1]) &&
        [f[:grain], t[:grain]].include?(args[2].to_sym)
    end
  when "time_open_interval_after"
    vals.any? do |v|
      i = v.dig(:Time, :Interval)
      next false unless i && i[:from] && i[:to].nil?
      f = leaf(i[:from])
      wc(f[:value]) == dt_parts(args[0]) && f[:grain] == args[1].to_sym
    end
  when "time_open_interval_before"
    vals.any? do |v|
      i = v.dig(:Time, :Interval)
      next false unless i && i[:from].nil? && i[:to]
      t = leaf(i[:to])
      wc(t[:value]) == dt_parts(args[0]) && t[:grain] == args[1].to_sym
    end
  when /\Ano_/ then entities.empty?
  else raise "unknown check #{check}"
  end
end

results = Hash.new { |h, k| h[k] = { pass: 0, fail: [] } }
CASES.each do |c|
  check = c["check"]; args = c["args"]; text = c["text"]
  unless text.is_a?(String)
    results[c["file"]][:fail] << { text: text.inspect, note: "extraction problem" }
    next
  end
  if check == "time_interval_with_context"
    ref = CTX.fetch(args[0][1])
    args = args[1..]
  elsif check.start_with?("time") || check == "no_time"
    ref = REF
  end
  kwargs = { locale: "en", dims: [DIM.fetch(check)] }
  kwargs[:reference_time] = ref if ref
  begin
    entities = Duckling.parse(text, **kwargs)
    if match?(check, args, entities)
      results[c["file"]][:pass] += 1
    else
      results[c["file"]][:fail] << { text: text, args: args, got: entities.map { |e| e[:value] }.first(2) }
    end
  rescue => e
    results[c["file"]][:fail] << { text: text, error: "#{e.class}: #{e.message}" }
  end
end

total_pass = results.values.sum { |r| r[:pass] }
total_fail = results.values.sum { |r| r[:fail].size }
results.sort.each do |file, r|
  puts format("%-35s pass %4d  fail %3d", file, r[:pass], r[:fail].size)
end
puts format("TOTAL %29s pass %4d  fail %3d", "", total_pass, total_fail)

if total_fail > 0
  puts "\n=== FAILURES ==="
  results.sort.each do |file, r|
    next if r[:fail].empty?
    puts "\n--- #{file} ---"
    r[:fail].each { |f| puts f.inspect }
  end
end
```

</details>

## Notes

- Verified on main @ `a6ee4f6` (v0.4.7). Related: #100 — the Data-object API proposal cites this corpus run as empirical backing for the tagged shape it wraps.
- Upstream pin at extraction time: wafer-inc/duckling @ `c96b068`.

cpb:
Our time tests mirrored the upstream EN corpus by hand, partially. The
`tests/*_corpus.rs` files in wafer-inc/duckling are the full ported Haskell
corpus, and they pin exactly the behaviours the unified `serde_magnus` shape
relies on: per-dim tagged hashes, grain symbols, real `Time` leaves,
open-interval `nil` bounds, `Measurement` `Value`/`Interval` discrimination,
and quantity products.

This extracts every `check_*` call site into a checked-in fixture, renders
committed minitest files from it, and replays the lot through `Duckling.parse`:
**1,656 upstream cases plus 6 local additions, as 663 test methods across 26
files under `test/wafer/`**, in about 1.2 seconds.

### Pipeline

Two stages, split at the network boundary:

```
wafer-inc/duckling tests/*_corpus.rs
  │  script/extract_wafer_corpus.rb        network — rake corpus:refresh
  ▼
test/fixtures/wafer_corpus.json + wafer_corpus_local.json
  │  script/generate_corpus_tests.rb       offline — rake corpus:generate
  │  script/templates/corpus_test.rb.erb
  ▼
test/wafer/<dimension>/<check>_test.rb     26 files, committed
```

`corpus:refresh` invokes `corpus:generate`, so one command still does a whole
update. Keeping the second stage offline and deterministic is what lets
`test/wafer_corpus_generation_test.rb` detect a stale or orphaned tree during
an ordinary `rake test` — no network, no subprocess, no CI change.

### Why generated, rather than built at load time

The first draft built its methods with `define_method` over the fixture. It
passed, but it contained no `def test_*` anywhere, and three things followed:

- **`bin/test <file>:<line>` could not reach any of it.** That script resolves
  a file:line ref by scanning for `/^\s*def\s+(test_\w+)/`, so every line of
  the corpus suite answered "no test method found" — the whole
  `bin/worktree heal-reproduce` contract was unavailable for 1,656 cases.
- **No test name was greppable.** The name in a failure message existed
  nowhere in the repo as text.
- **Every case shared one backtrace line.**

Generated files fix all three, and read like the Rust they came from:

```ruby
# time_corpus.rs:230
def test_time_today
  assert_time_naive "today", [2013, 2, 12, 0, 0, 0], :day
  assert_time_naive "at this time", [2013, 2, 12, 0, 0, 0], :day
end
```

One file per (dimension, check). The 659 upstream groups never mix check
kinds, so no `#[test] fn` is split across two files; each becomes one method
under the name upstream gave it. Class names are flat (`WaferTimeNaiveTest`)
rather than nested — a `Wafer::Time` module would shadow `::Time` inside the
very files whose subject is time values.

Assertions are fail-fast within a method, as upstream's Rust `assert!` is,
which is what makes the line number meaningful. Each matcher message carries
the text, the expectation, and the entities actually returned.

### Matcher semantics

`test/support/wafer_matchers.rb` mirrors the upstream Rust checkers. The rule
that costs the most to get wrong is measurement intervals: upstream accepts an
`Interval` when **either** bound matches value and unit ("matching original
test behavior"). A `Value`-only matcher reports 93 false failures on phrasings
such as "between 10 and 20 dollars". Also carried over verbatim: the 0.01
float tolerance, wall-clock component comparison, grain accepted on either
interval leg, and `.any?` over the returned entities — the corpus states that a
reading is present, not that it is the only one.

### What is left out

- Call sites behind a `//`. `time_corpus.rs` keeps 24, for two different
  reasons: 22 need `with_latent`, which no upstream checker passes, and 2
  (`diffCorpus`) need a third reference time, 2013-02-15 04:30 −02:00.
- `tests/pending_corpus.rs` — upstream knows those fail.

### Local additions

The crate's `by ` rule builds a *closed* interval `[now, named]`, so both
bounds are present even though only one is stated in the text. Consumer code
telling a deadline from a stated range has to compare `from` against the
reference time it passed in. Upstream pins that mechanism for
`by <clock time>` and `by <end of period>`, but never for a named day, so
`test/fixtures/wafer_corpus_local.json` adds `by <named day>` — `"by Friday"`,
`"by Friday July 3, 2026"`, `"by the 15th"` — where the `to` bound is a
`Naive` calendar day at `:second` grain. It is a separate file because the
refresh task rewrites the generated one wholesale.

### Notes

- `docs/wafer-corpus.md` covers the pipeline, the fixture schema, the matcher
  rules, the three reference times, and the deadline distinction.
- `NOTICES` covers the derived corpus data under the same BSD-3-Clause notice
  as the linked crate. Neither fixture, nor `script/`, nor `test/wafer/` is
  part of the packaged gem — `test/gemspec_files_test.rb` guards that.
- Everything under `test/wafer/` is generated. Hand edits fail the staleness
  test and are overwritten by the next `corpus:generate`.

Closes #157


script/generate_corpus_tests.rb
```diff
@@ -0,0 +1,203 @@
...
+        [expected.fetch("value"), expected.fetch("unit"), expected.fetch("product")].map(&:inspect)
+      when "time_interval"
+        [expected.fetch("from").inspect, expected.fetch("to").inspect, expected.fetch("grain").to_sym.inspect]
+      when "time_interval_with_context"
+        [expected.fetch("from").inspect, expected.fetch("to").inspect, expected.fetch("grain").to_sym.inspect,
+          "context: #{kase.fetch("context").inspect}"]
+      else [expected.fetch("value").inspect]
+      end
+
+    "#{ASSERTIONS.fetch(check)} #{([text.inspect] + literals).join(", ")}"
```
cpb, line 214: **Locale-dependent output — the staleness test false-fails on US-ASCII hosts.** `text.inspect` (and every `.map(&:inspect)` above) is sensitive to `Encoding.default_external`: when it isn't UTF-8 (`LANG=C`, the default in minimal containers like `ruby:*-slim` and some dev shells), `inspect` escapes non-ASCII instead of emitting it raw. Demonstrated on this branch:

```
$ LANG=C LC_ALL=C bundle exec ruby -Ilib -Itest test/wafer_corpus_generation_test.rb
WaferCorpusGenerationTest#test_generated_files_are_current:
  test/wafer/amount_of_money/money_test.rb is stale. Run `bundle exec rake corpus:generate`.
-    assert_money "10\u00A2", 10.0, "cent"
+    assert_money "10¢", 10.0, "cent"
```

So on a US-ASCII host `rake test` goes red on a correctly-generated tree — and worse, a dev who follows the error's advice rewrites money/ordinal/temperature into the escaped form, which then fails the staleness test for everyone on UTF-8 hosts. Locale ping-pong. The explicit `encoding: "UTF-8"` on the file reads/writes fixed byte I/O, but the locale leaks in at literal *generation*. This contradicts the "same fixtures always produce the same bytes" guarantee the staleness test relies on.

Alternatives:
1. Pin the encoding around rendering: save/set/restore `Encoding.default_external = Encoding::UTF_8` inside `CorpusTests.render_all` — one place, fixes both the CLI and the in-process staleness render (verified locally: `inspect` then emits raw UTF-8 under `LANG=C`).
2. Emit ASCII-only literals deterministically: post-process with e.g. `.gsub(/[^\x00-\x7F]/) { |c| format("\\u%04X", c.ord) }` — under US-ASCII `inspect` already escaped, under UTF-8 this escapes, so both hosts converge to identical bytes (must match `inspect`'s uppercase 4-digit format exactly).
3. Compare parsed Ruby values instead of bytes in the staleness test — weaker; loses the byte-pinning that makes hand-edits detectable. Not recommended.

script/generate_corpus_tests.rb
```diff
@@ -0,0 +1,203 @@
...
+  def file_stem(dimension, check)
+    check.delete_prefix("#{dimension}_")
+  end
+
+  def render(template, dimension, check, cases)
+    groups = cases.group_by { |kase| kase.fetch("group") }
+      .map { |name, group_cases| build_group(name, group_cases) }
+      .sort_by { |group| group.fetch(:sort_key) }
+
+    duplicates = groups.map { |group| group.fetch(:name) }.tally.select { |_, count| count > 1 }
```
cpb: **Dead guard.** `groups` is built from `cases.group_by { |kase| kase.fetch("group") }`, so group names are unique by construction — `tally.select { |_, count| count > 1 }` is always empty and this `raise` can never fire. Worse, the collision it wants to catch (two distinct upstream `#[test] fn`s with the same name landing in one `(dimension, check)` bucket, e.g. the same fn name in two corpus files) would be silently *merged* by that same `group_by` into a single method whose origin comment names only the first file — the exact failure mode the guard exists to prevent, with the guard giving false confidence. (Verified no group name spans files at c96b068, so this is latent today.)

Alternatives:
1. Check before grouping: `cases.group_by { |k| k.fetch("group") }.select { |_, cs| cs.map { |c| c.fetch("file") }.uniq.size > 1 }` and raise on that.
2. Group by `[file, group]` pairs and raise when any name appears more than once across pairs.
3. Delete the dead check and rely on minitest's duplicate-method behavior (last definition wins — silent, so prefer 1 or 2).

script/generate_corpus_tests.rb
```diff
@@ -0,0 +1,203 @@
...
+    raise "Duplicate method names in #{dimension}/#{check}: #{duplicates.keys.inspect}" if duplicates.any?
+
+    source = cases.first.fetch("file").end_with?(".rs") ? "test/fixtures/wafer_corpus.json" : "test/fixtures/wafer_corpus_local.json"
```
cpb: **Provenance inferred from a data field rather than the loader's own knowledge.** `cases.first.fetch("file").end_with?(".rs")` decides which fixture a bucket came from, but `annotate` already knows the fixture at load time. A hand-written local case whose `file` happens to end in `.rs` (a natural choice when mimicking upstream layout) would be mislabeled as upstream-extracted in the generated header, and `provenance` (line 131 has the same pattern) would then attach the extracted fixture's sha to a local file — a misleading trail exactly where provenance matters most.

Alternatives:
1. Thread the fixture name through `annotate`'s merge — it already merges `"dimension"`; add `"fixture" => name` — and branch on that here and in `provenance`.
2. Or key `FIXTURES` by name with an explicit provenance label per fixture, and look it up instead of re-deriving it from case data.

AGENTS.md
```diff
@@ -31,9 +31,14 @@ If you notice this file describing a not-yet-built piece as current, or vice ver
...
 | `test/fixtures/tz/` | `zic` source for the three synthetic zones (`Fixture/NegativeDst`, `Fixture/HalfHourGap`, `Fixture/LateGap`) plus the minimal `iso3166.tab`/`zone1970.tab` a zoneinfo directory needs to be recognized. Compiled at test time, never committed as binaries. |
 | `test/fixtures/zoneinfo-overrides/` | `zic` source for zones that `bin/build-stale-zoneinfo` compiles *over* a copy of the host's zoneinfo directory, rolling named zones back to earlier rules. Separate from `test/fixtures/tz/` on purpose: anything in that directory becomes a fixture-zone identifier, and the fixture datasource must expose only its own three. |
+| `test/fixtures/wafer_corpus.json` | Every `check_*` call site of the wrapped crate's own EN corpora (`tests/*_corpus.rs`), as 1,656 expectation records. Generated by `script/extract_wafer_corpus.rb` via `rake corpus:refresh` — never hand-edited. BSD-3-Clause derived data (see `NOTICES`). Read by `test/duckling_wafer_corpus_test.rb`; see [docs/wafer-corpus.md](docs/wafer-corpus.md). |
```
cpb: **Stale reference.** "Read by `test/duckling_wafer_corpus_test.rb`" — that file doesn't exist in this branch. It's a leftover from the first commit's load-time design, before the switch to generated `test/wafer/` files. Since this file's own preamble asks to be kept honest about what exists, this row will mislead the next reader (or agent) looking for the consumer of the fixture.

Alternatives:
1. Point at the real consumers: "Read by the generated `test/wafer/` tree (via `script/generate_corpus_tests.rb`); guarded by `test/wafer_corpus_generation_test.rb`."
2. Or drop the consumer clause and let the `test/wafer/` row two lines below carry it.

script/extract_wafer_corpus.rb
```diff
@@ -0,0 +1,271 @@
...
+  # True when the call site sits behind a `//` on its own line. The corpus
+  # has no block comments.
+  def commented_out?(src, offset)
+    line_start = src.rindex("\n", offset)
+    line_start = line_start ? line_start + 1 : 0
+    src[line_start...offset].include?("//")
+  end
+
+  def enclosing_test(src, offset)
+    src[0...offset][/.*^\s*fn\s+(\w+)\s*\(/m, 1] || "unknown"
```
cpb: **Silent `"unknown"` fallback.** If this regex ever stops matching (upstream reformats, a call site appears outside any `fn`), every affected case gets group `"unknown"`, and the generator's `group_by` then merges them all into one `def unknown` method — and since the generator's duplicate-name guard is dead (see my comment there), nothing raises anywhere. The extraction would look successful while quietly collapsing cases. `context_before` in this same file raises on the analogous failure; this should fail just as loudly.

Alternatives:
1. `|| raise("No enclosing fn for check_* call at offset #{offset}")` — consistent with `context_before` two methods up.
2. Or collect unmatched offsets and `warn` + include them in the summary line, if you'd rather a refresh succeed with visible degradation than abort.

script/extract_wafer_corpus.rb
```diff
@@ -0,0 +1,271 @@
...
+
+  RUST_ESCAPES = {"n" => "\n", "t" => "\t", "r" => "\r", "0" => "\0", "\\" => "\\", "\"" => "\"", "'" => "'"}.freeze
+
+  module_function
+
+  # Rust string literals in the corpus are one-line and use only simple
+  # escapes plus \u{...}.
+  def unquote(literal)
+    literal[1..-2]
+      .gsub(/\\u\{([0-9a-fA-F]+)\}/) { [$1.to_i(16)].pack("U") }
```
cpb: **Escape decoding order is wrong for `\\u{...}`, and the literal grammar is narrower than Rust's.** The `\u{...}` gsub runs before the backslash gsub, so a Rust literal containing an escaped backslash followed by `u{` (`"\\u{41}"` — i.e. the literal text `\u{41}`) decodes to `A` instead of `\u{41}`. Related latent gaps: `\xNN` (valid in Rust strings) isn't in `RUST_ESCAPES` and would survive as the literal text `xNN`, and `parse_expectation` only accepts plain `\d+` / `\d+\.\d+` — Rust's `1_000`, `1e5`, and `1.0f64` forms would raise. None of these appear in the corpus at c96b068 (verified: only simple escapes and `\u{...}`), so this is latent — but a refresh that adds one fails confusingly or, for the ordering bug, silently corrupts the text.

Alternatives:
1. Decode in a single left-to-right pass, e.g. `gsub(/\\(u\{([0-9a-fA-F]+)\}|x([0-9a-fA-F]{2})|(.))/m)` dispatching on which arm matched — one pass can't re-match its own output.
2. Or handle `\\\\` first, then `\u{...}`, then the rest — order matters more than completeness here.
3. For numerics, extend `parse_expectation`'s patterns to strip `_` and suffixes before `Integer`/`Float`.

script/extract_wafer_corpus.rb
```diff
@@ -0,0 +1,271 @@
...
+    raise "No make_*_context() binding before offset #{offset}" unless name
+    name
+  end
+
+  # True when the call site sits behind a `//` on its own line. The corpus
+  # has no block comments.
+  def commented_out?(src, offset)
+    line_start = src.rindex("\n", offset)
+    line_start = line_start ? line_start + 1 : 0
+    src[line_start...offset].include?("//")
```
cpb: **`include?("//")` isn't string-aware.** A `//` inside a string literal earlier on the same line (e.g. `let u = "https://…"; check_url(…)`) would mark a live call site as commented out and silently drop the case from the fixture. Note the naive fix — anchoring on `^\s*//` — would introduce the opposite bug, since a call after a *mid-line* `//` genuinely is commented out and would then be extracted. Today's corpus is one call per line with nothing before it (verified at c96b068), so this is latent.

Alternatives:
1. Reuse the string-skipping scan from `read_call`/`split_args` to walk the line prefix and look for `//` outside string literals — the file already has exactly the state machine needed.
2. Or assert the precondition loudly: `raise` if the line prefix contains a `"` before any `//`, turning a silent drop into a refresh-time error.

cpb (commented):
Adversarial review from a worktree at 4d5a3c5 (fresh clone of upstream at the pinned sha, full suite run, re-extraction diffed against the committed fixture). The load-bearing claims all check out — verified independently rather than from the description:

**Verified solid**
- Re-ran `script/extract_wafer_corpus.rb` against wafer-inc/duckling@c96b068: **byte-identical fixture**. An independent grep-based recount reconciles 1,656 active call sites (1,705 `check_*` tokens − 26 checker definitions − 24 commented-out − the `check_quantity_impl` helper nuance).
- Matcher semantics read true against the upstream Rust checkers: either-bound measurement intervals (`check_money`/`check_quantity_impl`), grain-on-either-leg, 0.01 float tolerance, `naive_local()` wall-clock comparison, exact duration/ordinal equality, and `expected_product: None ⇒ any product passes` for plain `check_quantity` — all mirrored correctly in `wafer_matchers.rb`.
- 663 `def test_` across 26 files; 659 upstream groups, none mixing check kinds, no group name spanning files; the 24 commented-out call sites are indeed 22 latent + 2 diffCorpus.
- `bin/test test/wafer/time/naive_test.rb:14` resolves to and runs exactly `test_time_today` — the generation-over-`define_method` rationale holds.
- `Instant` leaves carry their `FixedOffset` into the Ruby `Time` (`patch_time_point` in `ext/duckling/src/lib.rs`), so `wall_clock` is host-TZ-independent. Suite green locally (666 runs incl. 3 capability tests, ~1.3s).
- The gemspec allow-list keeps `script/`, `test/`, and the fixtures out of the packaged gem.

**Issues left inline**: one demonstrable portability bug (locale-dependent generation false-fails the staleness test on US-ASCII hosts), one dead guard, one stale doc reference, and four latent extractor/generator robustness notes. Each comment includes alternatives.

script/generate_corpus_tests.rb
cpb: Fixed in 584c187, alternative 1. `CorpusTests.render_all` now wraps rendering in `with_utf8_external`, which saves, sets, and restores `Encoding.default_external`.

Reproduced your failure first, then confirmed the fix on both sides:

- Before: `LANG=C` generation emitted `"10¢"`, and the staleness test went red on a correct tree.
- After: `LANG=C LC_ALL=C bundle exec rake test` is green — 850 runs, 0 failures — and `LANG=C LC_ALL=C bundle exec rake corpus:generate` leaves `test/wafer/` byte-identical. So both hosts converge on the literal-UTF-8 form, and the ping-pong is gone.

A UTF-8 host returns before the assignment, so CI never touches the global. That also keeps `-w` quiet there: a non-UTF-8 host does see Ruby's own `warning: setting Encoding.default_external` twice per render, which seemed worth accepting over silencing it.

The explanation went to `docs/wafer-corpus.md` rather than staying inline, per the repo's comment convention — it now says determinism here means host-independent, not merely repeatable.

---

script/generate_corpus_tests.rb
cpb: Fixed in 584c187, alternative 1. You're right that the old check was tautological — it tallied names drawn from a `group_by` on those same names.

`render` now groups once and inspects the files each group spans:

```ruby
spanning = grouped.select { |_, group_cases| files_in(group_cases).length > 1 }
```

A name cannot repeat inside one Rust file, so spanning files is the whole test. The message names the offender and both files:

```
Group name spans corpus files in numeral/numeral: test_dup (a_corpus.rs, b_corpus.rs)
```

Pinned by `test_render_refuses_a_group_name_that_spans_corpus_files`, with `test_render_accepts_distinct_group_names_from_one_file` on the other side so the guard can't pass by rejecting everything.

---

script/generate_corpus_tests.rb
cpb: Fixed in 584c187, alternative 1. `annotate` now takes the fixture name and merges `"fixture" => name` alongside `"dimension"`, so both call sites read the loader's own knowledge instead of re-deriving it:

- `source:` is `"test/fixtures/#{cases.first.fetch("fixture")}"`.
- `provenance` compares against a new `EXTRACTED_FIXTURE` constant rather than testing for a `.rs` suffix.

`test_provenance_labels_a_local_case_by_its_fixture` pins exactly your scenario — a local case whose `file` is `time_corpus.rs` — and it no longer picks up the extracted fixture's sha.

---

AGENTS.md
cpb: Fixed in 584c187, alternative 1. The row now reads:

> Rendered into the generated `test/wafer/` tree by `script/generate_corpus_tests.rb`, and pinned there by `test/wafer_corpus_generation_test.rb`

`grep -rn duckling_wafer_corpus_test` across the repo now returns nothing, so that was the last reference.

Two related AGENTS.md updates in the same commit, since the file asks for them: a directory-layout row for the new `test/wafer_corpus_extraction_test.rb`, and a Test-guide bullet covering both pipeline guards and the locale-independence property.

---

script/extract_wafer_corpus.rb
cpb: Fixed in 584c187, alternative 1 — it raises, matching `context_before` two methods up:

```ruby
raise "No enclosing fn for the check_* call at offset #{offset}" unless name
```

Chose failing loudly over the `warn`-and-continue option because of the interaction you spotted: the generator would merge every affected case into one method, so a degraded refresh produces a fixture that looks fine and quietly loses cases. Now that the generator's collision guard actually fires, a `"unknown"` bucket spanning files would also trip that — but only if the cases came from different files, so this remains the real guard.

Pinned by `test_enclosing_test_raises_when_there_is_no_enclosing_fn`, with `test_enclosing_test_names_the_enclosing_fn` for the normal path.

---

script/extract_wafer_corpus.rb
cpb: Fixed in 584c187 — alternative 1 for the ordering, plus alternative 3 for the numerics.

Escapes now decode in a single left-to-right pass over one alternation, so no arm can re-read another's output:

```ruby
ESCAPE = /\\(?:u\{([0-9a-fA-F]+)\}|x([0-9a-fA-F]{2})|(.))/m
```

`\xNN` joins the grammar as a codepoint rather than a byte, since Rust caps it at `0x7F` in string literals.

`parse_expectation` delegates its numeric arms to a new `parse_number` accepting `_` separators, exponents, and `i*`/`u*`/`f32`/`f64` suffixes, choosing Integer or Float from the fraction, exponent, or float suffix. Unrecognized arguments still raise.

Verified the ordering bug directly — `"\\u{41}"` decoded to `A` before, and now stays the literal `\u{41}` — and pinned the whole set in `test_unquote_*` / `test_parse_expectation_*`, including that `-2` stays an Integer and `98.6` a Float so the corpus's existing forms are unchanged.

One caveat on scope: I did not re-run `corpus:refresh` against upstream, since that reaches a repository outside this session's scope. The byte-identity I can show is that `corpus:generate` leaves `test/wafer/` untouched, which exercises the fixture but not the extractor. The extractor's own equivalence rests on the unit tests above rather than on a fresh re-extraction — worth re-confirming with a real refresh before the next corpus bump.

---

script/extract_wafer_corpus.rb
cpb: Fixed in 584c187, alternative 1. `commented_out?` now delegates to a `comment_index` scanner that tracks string literals and escapes, mirroring the state machine in `read_call`/`split_args`:

```ruby
elsif char == "/" && line[index + 1] == "/"
  return index
```

Your note about the naive `^\s*//` fix is what made me keep it a scan rather than an anchor — three tests pin both directions: `check_url(` after `let u = "https://x";` reads as live, while both `    // check_url(` and `foo(); // check_url(` still read as commented out.

Not handled, and unchanged from before: Rust raw strings (`r"..."`), char literals, and block comments. `read_call`/`split_args` don't handle those either, so this stays consistent with the rest of the extractor rather than being a new gap — and the header comment still records the no-block-comments assumption.

---

cpb:
...

test/wafer_corpus_generation_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+# Guards the two fixtures under test/fixtures/ and the tree generated from
+# them under test/wafer/. Deliberately outside test/wafer/, so that every
+# *_test.rb in there is generated with no hand-written file mixed in — which
+# is what makes the orphan check below a simple set comparison.
+#
+# The generator renders to strings, so this compares against disk in-process:
+# no subprocess, no network, no git. See docs/wafer-corpus.md.
+class WaferCorpusGenerationTest < Minitest::Test
+  def test_generated_files_are_current
```
cpb, line 14: **The locale-independence property is fixed and documented, but no test pins it.** `CorpusTests.render_all` renders identical bytes under any `Encoding.default_external` (verified by hand: `LANG=C LC_ALL=C bundle exec rake corpus:generate` leaves the tree byte-identical, and the full suite is green under `LANG=C`). But if `with_utf8_external` is ever removed or broken in a refactor, every UTF-8 host — which is all of CI — stays green; only a US-ASCII host would catch the regression. That is exactly the host this suite's guards otherwise free us from needing.

A cheap in-process pin, no subprocess or CI leg required:

```ruby
def test_rendering_is_independent_of_the_host_locale
  previous = Encoding.default_external
  Encoding.default_external = Encoding::US_ASCII
  rendered = CorpusTests.render_all
ensure
  Encoding.default_external = previous
end
```

…then assert `rendered` matches the on-disk tree byte-for-byte (same comparison `test_generated_files_are_current` already does). The fixture carries real non-ASCII (`10¢`, an em dash), so the two encodings genuinely diverge without the pin.

cpb (commented):
Three non-blocking improvement opportunities from an independent pass over the head commit (suite re-run green under UTF-8, LANG=C, and non-default TZ; re-extraction against upstream@c96b068 byte-identical; Value-only measurement matcher measured at exactly the documented 93 false failures). Inline: one unpinned property worth a test, one assumption worth naming, one minor efficiency note.

script/generate_corpus_tests.rb
```diff
@@ -0,0 +1,242 @@
...
+        .sort_by { |(dimension, check), _| [dimension, check] }
+        .to_h do |(dimension, check), cases|
+          path = File.join(OUTPUT_DIR, dimension, "#{file_stem(dimension, check)}_test.rb")
+          [path, render(template, dimension, check, cases)]
+        end
+    end
+  end
+
+  # Renders the same bytes whatever the host's locale is. `String#inspect`
+  # would otherwise escape non-ASCII on a US-ASCII host. Process-global state,
```
cpb, line 71: **Worth naming the assumption this safety rests on.** "Process-global state, hence the ensure" covers restoration, but the larger hazard is concurrency: assigning `Encoding.default_external` is only safe here because the suite never parallelizes (verified: no `parallelize` anywhere in `test/`). If minitest parallel execution is ever switched on, the staleness test could flip another thread's encoding mid-read.

Alternatives:
1. One more sentence in this comment, e.g. "Safe because the suite never runs tests in parallel; if that changes, guard this with a mutex."
2. Or take a small mutex around the assignment now, so the property doesn't depend on a convention enforced nowhere.

Rakefile
```diff
@@ -161,6 +161,32 @@ task :native_gem, [:platform] do |_t, platform:|
...
 
+# The EN corpora vendored into test/fixtures/wafer_corpus.json come from this
+# repository. See docs/wafer-corpus.md.
+UPSTREAM_CORPUS_REPO = "https://github.com/wafer-inc/duckling"
+CORPUS_CLONE_DIR = "tmp/wafer-duckling"
+
+namespace :corpus do
+  desc "Re-extract test/fixtures/wafer_corpus.json from wafer-inc/duckling, then regenerate (e.g. `rake 'corpus:refresh[c96b068]'`)"
+  task :refresh, [:ref] do |_task, ref: nil|
+    rm_rf CORPUS_CLONE_DIR
```
cpb, line 172: **Minor: every refresh re-clones from scratch.** `rm_rf` + a fresh blobless clone is simple and always correct, and blobless clones are cheap, so this is fine to keep — but a repeat refresh could instead reuse the clone:

```ruby
if File.directory?(CORPUS_CLONE_DIR)
  sh "git", "-C", CORPUS_CLONE_DIR, "fetch", "--filter=blob:none", "origin"
else
  sh "git", "clone", "--filter=blob:none", UPSTREAM_CORPUS_REPO, CORPUS_CLONE_DIR
end
```

That also drops the network cost to roughly the delta when pulling a newer upstream HEAD. No change needed if the always-clean-slate property is deliberate.

test/wafer_corpus_generation_test.rb
cpb: Added in e1bf77c, essentially your sketch — `test_rendering_does_not_depend_on_the_host_locale` renders under `Encoding::US_ASCII` and reuses the same byte comparison `test_generated_files_are_current` does.

Mutation-tested it rather than assuming, since a guard that can't fail is the thing being fixed here. With `with_utf8_external` neutered to a bare yield, on this UTF-8 host:

```
1) Failure:
test/wafer/amount_of_money/money_test.rb renders differently on a US-ASCII host.
See with_utf8_external in script/generate_corpus_tests.rb.
9 runs, 1 failures
```

Pin restored: 9 runs, 0 failures. So it catches the regression on exactly the hosts that previously couldn't.

One wrinkle worth naming: the encoding assignments warn under the suite's `-w`, which would have added noise to every run on every host. The helper silences `$VERBOSE` around them, following the same pattern `test_helper.rb` already uses for its circular capability requires. Restore order in the `ensure` matters — encoding first, `$VERBOSE` second — or the restore warns on its way out.

---

script/generate_corpus_tests.rb
cpb: Took alternative 1 in e1bf77c — the inline comment now ends "and safe only because the suite runs single-threaded" — but I want to push back on alternative 2, because I think the mutex would be worse than nothing here.

A lock around the assignment serializes two *renders* against each other. It does not scope the encoding change, because `Encoding.default_external` is process-global: for the duration of a render, every thread sees UTF-8, whether or not it ever touches `CorpusTests`. So under parallel minitest the hazard you describe — a thread reading with an encoding it did not choose — survives the mutex untouched. What the lock would change is that the race stops being obvious, which is the wrong direction for a property whose whole defence is that it is currently easy to reason about.

So: the assumption is named where it can be seen, and `docs/wafer-corpus.md` now spells out why the lock is not the fix and that turning on parallelization is what would need care. If the suite ever does parallelize, the real options are rendering in a subprocess or dropping to alternative 2 from your earlier comment (deterministic ASCII-only escaping, no global at all).

---

Rakefile
cpb: Kept the clean slate, and took your "no change needed if it's deliberate" as the prompt to make it visibly deliberate — e1bf77c records the reason where the `rm_rf` is.

Two things make reuse worse than the saved bandwidth here. `git fetch` does not move the checkout, so with no `ref` argument a reused clone would keep extracting from whatever HEAD it was left at — the refresh would quietly stop refreshing, and `upstream_sha` would faithfully record the stale sha, so the fixture would look correct. A reused directory can also be dirty from a previous poke, and then the recorded sha describes something other than what was extracted. Both failures produce a plausible-looking fixture, which is the worst shape for this file given it is reviewed by diff.

The blobless clone keeps the cost of that guarantee low enough that I would rather pay it. Worth revisiting if refresh ever becomes frequent enough for the network cost to matter — at that point fetch plus an explicit `reset --hard` to the resolved ref would be the safe form, not fetch alone.

---

script/generate_corpus_tests.rb
cpb: Verified in e1bf77c — the comment now names the single-threaded assumption, and docs/wafer-corpus.md spells out the rest. Your pushback on the mutex is right: the encoding is process-global, so a lock would serialize renders against each other while leaving every other thread exposed during the window — the race would just get quieter. Naming the assumption (alternative 1) plus the docs note is the correct outcome here. Resolving.

Rakefile
cpb: Verified in e1bf77c — the `rm_rf` now carries its rationale, which is all my note asked for. The reasoning is sound and worth having on record: `git fetch` doesn't move the checkout, so a reused clone with no `ref` would keep extracting from a stale HEAD while `upstream_sha` faithfully recorded it — a plausible-looking fixture that is wrong, the worst shape for a file reviewed by diff. Clean-slate plus blobless clone is the right trade at this cadence. Resolving.

test/wafer_corpus_generation_test.rb
cpb: Verified in e1bf77c — confirming this resolution was earned. The test passes at HEAD, and I re-ran the mutation the commit message describes: with `with_utf8_external` neutralized to a plain `yield`, this test goes red on a UTF-8 host (US-ASCII render escapes `€` to `\u20AC` where the committed tree has the literal). Restored, green again — including the full suite under `LANG=C LC_ALL=C` (851 runs, 0 failures). The pin is now genuinely pinned.

